PR #18, opened 2018-02-01
(PDK-569) puppet resource, now displays type

tphoney:


lib/puppet/resource_api/glue.rb
```diff
@@ -4,11 +4,12 @@ module Puppet::ResourceApi
   class TypeShim
     attr_reader :values
 
-    def initialize(title, resource_hash)
+    def initialize(title, resource_hash, type)
       # internalize and protect - needs to go deeper
       @values        = resource_hash.dup
       # "name" is a privileged key
       @values[:name] = title
+      @values[:type] = type
```
DavidS: `@values` is used for property/parameter storage. Use a separate instance var to avoid confusing other parts of the machinery.

There are also no acceptance tests for this, I believe, so it'll require some manual testing.

tphoney: agreed @DavidS

tphoney: also code coverage

DavidS: This code is particularly hard to unit test, since it would require all of puppet around it to exercise. With proper acceptance testing (even manual), we can give this one a pass.

DavidS:
As discussed in the comments above, this gets a pass on coverage, because the Shim classes are glue code that is only exercised when running the full puppet stack. @tphoney did manual testing to confirm no regressions for this.

PR #21, opened 2018-02-15
(FM-6797) Add debug logging of current and target states

da-ar:


willmeek:
+1

tphoney:
👍

CmdrCody:
How can I prevent Sensitive data from leaking onto the Logs when running in debug mode?

I have providers which accept passwords as a parameter, and they are leaking password data when run in debug mode after converting them to using the new SDK/Resource API.

DavidS:
The `Sensitive` problem is now being tracked in https://tickets.puppetlabs.com/browse/PDK-1091

PR #23, opened 2018-02-21
(PDK-536) Proper datatype parsing and checking

DavidS:
This uses Puppet4+ pops to parse and check datatypes. This allows types
to use the full range of expressibility when declaring attributes.

### DIFF
--- a/lib/puppet/resource_api.rb
+++ b/lib/puppet/resource_api.rb
@@ -89,7 +89,7 @@ module Puppet::ResourceApi
             #   if type.instance?(v)
             #     return true
             #   else
-            #     inferred_type = Puppet::Pops::Types::TypeCalculator.infer_set(value)
+            #     inferred_type = Puppet::Pops::Types::TypeCalculator.infer_set(v)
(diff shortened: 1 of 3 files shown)


lib/puppet/resource_api.rb
```diff
@@ -99,12 +99,33 @@ def my_provider
...
             end
 
-            case options[:type]
-            when 'String'
+            type = Puppet::Pops::Types::TypeParser.singleton.parse(options[:type])
+            validate do |value|
+              return true if type.instance?(value)
+
+              inferred_type = Puppet::Pops::Types::TypeCalculator.infer_set(value)
+              if inferred_type.is_a? Puppet::Pops::Types::PStringType
```
thallgren: `TypeCalculator#infer_set` can be rather expensive and the `inferred_type` is not really of any use unless there's a mismatch so I'd suggest using:

    if value.is_a?(String)

instead and move the call to `#infer_set` down to where the mismatch error is generated.

DavidS: Thanks, that makes sense.

PR #37, opened 2018-03-20
(PDK-512) Add support for simple_get_filter

da-ar:


spec/puppet/resource_api/simple_provider_spec.rb
```diff
@@ -3,9 +3,9 @@
 
 RSpec.describe Puppet::ResourceApi::SimpleProvider do
   let(:context) { instance_double('Puppet::ResourceApi::BaseContext', 'context') }
-  let(:provider_class) do
+  let(:provider_class) do # here...
```
DavidS: there?

lib/puppet/resource_api.rb
```diff
@@ -11,11 +11,10 @@ def register_type(definition)
     raise Puppet::DevError, 'requires attributes' unless definition.key? :attributes
 
     definition[:features] ||= []
-    supported_features = %w[supports_noop canonicalize remote_resource].freeze
+    supported_features = %w[supports_noop canonicalize remote_resource simple_get_filter].freeze
     unknown_features = definition[:features] - supported_features
     Puppet.warning("Unknown feature detected: #{unknown_features.inspect}") unless unknown_features.empty?
 
-
```
DavidS: whitespace should be fixed in previous commit

lib/puppet/resource_api/simple_provider.rb
```diff
@@ -7,7 +7,12 @@ module Puppet::ResourceApi
   class SimpleProvider
     def set(context, changes)
       changes.each do |name, change|
-        is = change.key?(:is) ? change[:is] : (get(context) || []).find { |r| r[:name] == name }
+        is = if context.feature_support?('simple_get_filter')
+               change.key?(:is) ? change[:is] : (get(context, [name]) || []).first
```
DavidS: according to the spec, `get(context, [name])`  is allowed to return more than one resource, and the first is not necessarily the one we were asking for. This still needs the same `.find { |r| r[:name] == name }` as below.

PR #48, opened 2018-03-26
(PDK-882,PDK-883) validate only when needed

DavidS:


spec/acceptance/validation_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+      expect(status.exitstatus).to eq 0
+    end
+
+    it 'allows creating' do
+      output, status = Open3.capture2e("puppet apply #{common_args} -e \"test_validation{ new: prop => 2, param => 3 }\"")
+      expect(output.strip).to match %r{Test_validation\[new\]/ensure: defined 'ensure' as 'present'}
+      expect(status.exitstatus).to eq 0
+    end
+
+    it 'validates property' do
```
da-ar: new test for 'A mandatory property, that MUST NOT be validated on deleting.
```
output, status = Open3.capture2e("puppet apply #{common_args} -e \"test_validation{ foo: ensure => absent, prop => not_a_number }\"")
expect(output.strip).to match %r{Test_validation\[foo\]/ensure: undefined 'ensure' from 'present'}
expect(status.exitstatus).to eq 0
```

DavidS: I'd rather add a bunch of unit tests. The acceptance tests were primarily for the `@called_from_resource`, which I didn't need for PDK-883 in the end.

PR #55, opened 2018-04-06
(PDK-911) Fix handling of `ensure` values from symbols to strings

DavidS:
Tripping over puppet's behaviour for handling newvalue values, rsapi
would accidentally pass symbols for `Enum[absent, present]` ensures.

The unit and integration tests did not agree on whether `absent` and
`present` should be strings or symbols. Puppet itself is calling for
a string as a value for `Enum`s, and in the implementation those two
were only symbols because of a special case in the setup of the two.

This commit fixes this behaviour to always use string. This makes it
consistent with the rest of the API, and reduces the amount of noise
developers have to remember when working with values passed from us.

### DIFF
--- a/lib/puppet/resource_api.rb
+++ b/lib/puppet/resource_api.rb
@@ -85,7 +85,7 @@ module Puppet::ResourceApi
 
           # skip properties if the resource is being deleted
           next if definition[:attributes][:ensure] &&
-                  value(:ensure) == :absent &&
+                  value(:ensure) == 'absent' &&
                   options[:behaviour].nil?
 
           if value(name).nil? && !(type.instance_of? Puppet::Pops::Types::POptionalType)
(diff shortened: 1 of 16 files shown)


lib/puppet/resource_api.rb
```diff
@@ -210,7 +210,11 @@ def feature_support?(feature)
 
           case options[:type]
           when 'Enum[present, absent]'
-            Puppet::ResourceApi.def_newvalues(self, param_or_property, :absent, :present)
+            Puppet::ResourceApi.def_newvalues(self, param_or_property, 'absent', 'present')
```
tphoney: jeez 😞

tphoney:
👍 lets do eet

lib/puppet/resource_api.rb
DavidS: the real bummer is the next line. this here is only cosmetics

PR #66, opened 2018-04-13
Cache ~/.rvm for jruby jobs

cotsog:
Hey there,

I'm a Customer Support Engineer @ Travis CI and @DavidS wrote asking if it was possible to speed up the jruby jobs.

It's not great but here's my attempt. The idea is to cache the ~/.rvm directory and install jruby manually instead of using the `rvm:` key in the .travis.yml file.

Let's follow up in the e-mail you sent.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,18 @@ matrix:
   include:
     - rvm: 2.4.3
       env: PUPPET_GEM_VERSION='~> 5' # 5.5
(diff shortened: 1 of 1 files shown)


DavidS:
@cotsog brilliant, the test times are down to 3:02 (jruby 1.7), and 2:33 (jruby9k) from 4:39 and 6:52 respectively.

:heart: 

Thank you so much!

PR #72, opened 2018-04-25
(PDK-955) Expose type to context

da-ar:


lib/puppet/resource_api.rb
```diff
@@ -341,7 +341,7 @@ def call_provider(value); end
       end
 
       define_singleton_method(:context) do
-        @context ||= PuppetContext.new(definition[:name])
+        @context ||= PuppetContext.new(definition[:name], definition)
```
DavidS: This can be tightened up to only pass the definition, and have the name as a `Type` attr_reader.

lib/puppet/resource_api/simple_provider.rb
```diff
@@ -16,6 +16,8 @@ def set(context, changes)
 
         should = change[:should]
 
+        raise 'SimpleProvider cannot be used with a Type that is not ensurable' unless context.type.ensurable?
+
```
DavidS: Please pull the actual PDK-924 change out into its own commit. And then PR the `context.type` change as a separate PR, so it gets its own CHANGELOG entry. This is important enough to warrant that.

spec/puppet/resource_api/base_context_spec.rb
```diff
@@ -10,9 +10,13 @@ def send_log(log, msg)
   end
 
   subject(:context) do
-    TestContext.new('some_resource')
+    TestContext.new('some_resource', attributes)
```
DavidS: That's `definition`, not `attributes` now. (repeat for all occurrences below)

spec/puppet/resource_api/base_context_spec.rb
```diff
@@ -340,13 +344,28 @@ def send_log(log, msg)
...
 
     context 'when type does not support a feature' do
       let(:return_val) { false }
 
-      it { expect(described_class.new('supported_feature')).not_to be_feature_support('simple_get_filter') }
+      it { expect(described_class.new('supported_feature', attributes)).not_to be_feature_support('simple_get_filter') }
+    end
+  end
+
+  describe '.type' do
```
DavidS: Arguably this could be a stand-alone context for `...::BaseContext::Type`, only testing `Type.new(definition)`, and testing the `context.type` accessor further up.

@da-ar , you can leave it as is, if you don't think that it'll improve readability.

spec/puppet/resource_api/simple_provider_spec.rb
```diff
@@ -177,4 +183,25 @@ def delete(context, _name); end
...
+      { 'title' =>
+            {
+              is: is_values,
+              should: should_values,
+            } }
+    end
+
+    before(:each) do
+      allow(context).to receive(:updating).with('title').and_yield
+      allow(context).to receive(:feature_support?).with('simple_get_filter')
```
DavidS: Seeing that here reminds me that `feature_support?` could become `type.has_feature?` (c.f. `hash.has_key?`). Note that since we're already 1.0, we can't just delete `feature_support?` either way.

lib/puppet/resource_api.rb
```diff
@@ -377,6 +370,14 @@ def context
...
         end
       end
+
+      define_singleton_method(:context) do
+        @context ||= PuppetContext.new(definition)
+      end
+
+      def context
+        self.class.context
+      end
```
DavidS: ^^

lib/puppet/resource_api/base_context.rb
```diff
@@ -19,13 +24,8 @@ def failed?
...
 
   def feature_support?(feature)
-    supported = Puppet::Type.type(@typename).feature_support?(feature)
-    if supported
-      Puppet.debug("#{@typename} supports `#{feature}`")
-    else
-      Puppet.debug("#{@typename} does not support `#{feature}`")
-    end
-    supported
+    Puppet.info('DEPRECATED: Please use context.type.feature? instead.')
```
DavidS: Could use puppet's built-in `deprecation_warning(msg, key)` function: https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/puppet/functions/deprecation.rb#L28
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/logging.rb#L129

PR #74, opened 2018-05-02
(PDK-946) Passes ensure values to puppet as symbols.

da-ar:
Providers still deal with ensure values as strings, but puppet is always
passed symbols.


DavidS:
Apologies for not pointing at https://github.com/puppetlabs/puppet-resource_api/pull/62/files earlier!

DavidS:
I've added some changes to avoid the mungify changes. Still needs some more cleanup to revert the `name` change to mungify, maybe others? too late today.

PR #80, opened 2018-05-22
(PDK-988) restrain mungify from non-`puppet resource` workflows

DavidS:


lib/puppet/resource_api.rb
```diff
@@ -517,13 +517,30 @@ def self.try_mungify(type, value, error_msg_prefix)
       # fall through to default handling
     end
 
-    # a match!
-    return [value, nil] if type.instance?(value)
+    is_valid, error_msg = try_validate(type, value, error_msg_prefix)
```
da-ar: Possibly, have try_validate return value, error_msg? If it fails, return nil, error_msg.

lib/puppet/resource_api.rb
```diff
@@ -527,6 +541,18 @@ def self.try_mungify(type, value, error_msg_prefix)
...
+  # Validates the `value` against the specified `type`.
+  # @param type[Puppet::Pops::Types::TypedModelObject] the type to check against
+  # @param value the value to clean
+  # @param error_msg_prefix[String] a prefix for the error messages
+  # @raise [Puppet::ResourceError] if `value` is not of type `type`
+  # @private
+  def self.validate(type, value, error_msg_prefix)
+    is_valid, error_msg = try_validate(type, value, error_msg_prefix)
+
+    raise Puppet::ResourceError, error_msg unless is_valid
```
da-ar: if previous suggestion is followed `unless is_valid` would change here.

lib/puppet/resource_api.rb
DavidS: Doing it like this would obscure the fact that `try_validate` doesn't change the value. But I'll change it to only return `error_msg` or `nil`, which is enough information, and lines up better with `try_mungify`.

PR #82, opened 2018-05-24
(PDK-531) Support for composite namevars

da-ar:
Allows Types to provide a regular expression containing the namevar
attributes which make up the unique title.


DavidS:
:+1:

PR #87, opened 2018-05-31
(PDK-996) Provide better messaging when type cannot be resolved

da-ar:
Also contains:

Better messaging when provider.get does not return correctly
(maint) Pre-commit hook should now properly deal with stages changes

### DIFF
--- a/lib/puppet/resource_api.rb
+++ b/lib/puppet/resource_api.rb
@@ -170,7 +170,14 @@ module Puppet::ResourceApi
             end
           end
 
-          type = Puppet::Pops::Types::TypeParser.singleton.parse(options[:type])
+          begin
+            type = Puppet::Pops::Types::TypeParser.singleton.parse(options[:type])
+          rescue Puppet::ParseErrorWithIssue => e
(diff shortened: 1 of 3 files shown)


lib/puppet/resource_api.rb
```diff
@@ -262,17 +269,21 @@ def call_provider(value); end
...
+            end
+            context.type.namevars.each do |namevar|
+              if resource_hash[namevar].nil?
+                raise Puppet::ResourceError, "`#{name}.get` did not return a value for the `#{namevar}` namevar attribute"
+              end
             end
+            Puppet::ResourceApi::TypeShim.new(resource_hash, name, context.type.namevars, attr_def)
           end
-          Puppet::ResourceApi::TypeShim.new(resource_hash, name, context.type.namevars, attr_def)
+        rescue NoMethodError => e
```
DavidS: * This might hide/modify errors from the `get()` call itself. 
* This is relevant to PDK-917, but the commit is not tagged
* There are many more calls of `get()` relevant to PDK-917 that need checking
* I've updated pdk-917 to be more specific about what has to be checked

da-ar: Have removed this commit from the PR, will work on separately for PDK-917

PR #88, opened 2018-06-06
(PDK-917) Validates provider.get values against Type schema

da-ar:


spec/fixtures/test_module/.rubocop.yml
```diff
@@ -56,10 +56,6 @@ Style/TrailingCommaInArguments:
   Description: Prefer always trailing comma on multiline argument lists. This makes
     diffs, and re-ordering nicer.
   EnforcedStyleForMultiline: comma
-Style/TrailingCommaInLiteral:
-  Description: Prefer always trailing comma on multiline literals. This makes diffs,
-    and re-ordering nicer.
-  EnforcedStyleForMultiline: comma
```
DavidS: Please revert that

spec/acceptance/simple_get_filter_spec.rb
```diff
@@ -4,7 +4,7 @@
 require 'tempfile'
 
 RSpec.describe 'exercising simple_get_filter' do
-  let(:common_args) { '--verbose --debug --trace --modulepath spec/fixtures' }
+  let(:common_args) { '--verbose --debug --trace  --strict=error  --modulepath spec/fixtures' }
```
DavidS: There is a space too much here (and everywhere you copy&pasted this)

lib/puppet/resource_api.rb
```diff
@@ -304,6 +290,8 @@ def call_provider(value); end
                           my_provider.get(context).find { |h| namevar_match?(h) }
                         end
 
+        @warning_count = 0 # always reset when called
```
DavidS: Don't reset here. After 100 warnings for a type, there is no additional info that devs can get from this, and we don't want to kill production if someone slips in hitting this.

spec/acceptance/provider_validation_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+RSpec.describe 'a provider that returns data that does not match the type schema' do
+  let(:common_args) { '--verbose --trace  --modulepath spec/fixtures' }
+  let(:message) { 'Provider returned data that does not match the Type Schema for `provider_validation' }
+
+  describe 'when strict is default (warning)`' do
+    let(:prefix) { 'Warning:' }
+
+    it 'is raises the expected warnings' do
+      stdout_str, status = Open3.capture2e("puppet resource #{common_args} provider_validation")
```
DavidS: You can re-use the expectations here by wrapping the `describe`s in a loop like this:

```
[ ['warning', 'Warning: '], ['error', 'Error: Could not run:'], ...].each do |strict_value, prefix_value|
describe "when strict is set to #{strict_value}" do
  let(:common_args) { super() + " --strict=#{strict_value}" }
  let(:prefix) { prefix_value }
...
```

lib/puppet/resource_api/type_definition.rb
```diff
@@ -35,4 +35,56 @@ def feature?(feature)
...
+    namevars.each do |namevar|
+      if resource[namevar].nil?
+        raise Puppet::ResourceError, "`#{name}.get` did not return a value for the `#{namevar}` namevar attribute"
+      end
+    end
+
+    message = "Provider returned data that does not match the Type Schema for `#{name}[#{resource[namevars.first]}]`"
+
+    rejected_keys = check_schema_keys(resource) # removes bad keys
+    bad_values = check_schema_values(resource.dup)
```
DavidS: Is `check_schema` allowed to modify `resource` or not? If yes, do you need to `dup` here? If not, shouldn't this `dup` before `check_schema_keys`?

Also, now that the question has come up, you _might_ want to add a unit test making sure that `resource` doesn't get modified when checking.

PR #89, opened 2018-06-06
Improve unit tests of parameter and property results after register_type

DavidS:


DavidS:
some quick fixes in preparation of pdk-1007

PR #91, opened 2018-06-06
(PDK-1007) the namevar is a param

DavidS:
Based on https://puppet.com/docs/puppet/5.5/custom_types.html#namevar

### DIFF
--- a/lib/puppet/resource_api.rb
+++ b/lib/puppet/resource_api.rb
@@ -94,6 +94,17 @@ module Puppet::ResourceApi
         super(attributes)
       end
 
+      def to_resource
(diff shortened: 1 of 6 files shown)


lib/puppet/resource_api.rb
```diff
@@ -94,6 +94,17 @@ def type_definition
         super(attributes)
       end
 
+      def to_resource
+        to_resource_shim(super)
+      end
+
+      define_method(:to_resource_shim) do |resource|
+        # require'pry';binding.pry
```
da-ar: 🗑

PR #92, opened 2018-06-06
Enable randomised rspec testing

da-ar:


spec/puppet/resource_api_spec.rb
```diff
@@ -1162,22 +1161,25 @@ def set(_context, changes)
...
             it('set will be called with the correct structure') do
-              expect(instance.my_provider.last_changes).to eq('somename' => {
-                                                                is: { name: 'somename', test_string: 'canonfoo' },
-                                                                should: { name: 'somename', test_string: 'canonbar' },
-                                                              })
-            end
+              run_one = type.new(name: 'somename', test_string: 'foo')
+              run_one.flush
+              run_two = type.new(name: 'somename', test_string: 'bar')
+              run_two.flush
```
DavidS: shouldn't this go into a `before` block, instead of conflating the two examples?

PR #97, opened 2018-06-12
Enable parallel rspec testing

da-ar:


DavidS:
One thing: looking at https://codecov.io/gh/puppetlabs/puppet-resource_api/commit/e4f0017236d5bd0f1bd1b8b2a8666dfa833e20f5/build it seems to be uploading a result for each thread, instead for each cell. That might cause the extra processing delay on their end. Is this something we can fix?

DavidS:
Comparing recent master builds (4'30 to 6' build time) with the builds on this branch (3'20 and 8') the gains are not overwhelming. One of the prime reasons I believe is that the travis setup does not allow for loads of parallelization. 

As a downside this setup, output complexity both on travis and codecov is higher due to the interleaving of threads, and the lackluster tool support that does not create proper summaries.

I'm inclined to stop trying to get this in.

PR #103, opened 2018-06-21
Improve log_exception output from PuppetContext

da-ar:


spec/puppet/resource_api/puppet_context_spec.rb
```diff
@@ -19,12 +19,25 @@
...
     end
+
+    it 'will log message at error level and without trace' do
+      expect(Puppet::Util::Log).to receive(:create).with(level: :err, message: match('message'))
+      context.log_exception(exception, message: 'message', trace: false)
+    end
+
+    context 'when Puppet[:trace] is enabled' do
+      before(:each) do
+        Puppet[:trace] = true
```
DavidS: This will poison the settings for everyone else (especially if you also want to test `Puppet[:trace] = false`. Have you tried 
```
allow(Puppet).to recieve[:[].and_call_original
allow(Puppet).to recieve[:[].with(:trace).and_return(true)
```
?

PR #105, opened 2018-07-04
{WIP} (FM-7071) Adding custom facts for devices

tphoney:


tphoney:
Comments welcome, points of note. 

1.  Should the resource-api also load the facts, this would remove code from the module's implementation ?
2. return_custom_facts looks in Puppet[:pluginfactdest], should it be somewhere else ?
3. Should return_custom_facts be called as part of the facts function ?

README.md
```diff
@@ -175,6 +175,48 @@ The provider needs to specify the `remote_resource` feature to enable the second
...
+class Example_ResourceAPI_Fact
+  def self.add_fact(connection, facts)
+    facts['baz'] = 'foo'
+    facts
+  end
+end
+```
+and the fact lives here EG:
+```
+/opt/puppet/cache/devices/3750/facts.d/
```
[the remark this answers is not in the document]
DavidS: This also needs to be reevaluated under the recent conversations about recommended deployment modes for `puppet device` running.

spec/acceptance/device_spec.rb
```diff
@@ -103,6 +118,15 @@
       expect(status).to eq 0
     end
 
+    it 'has the custom fact "baz" fact set to "foo"' do
+      fact_dir = File.dirname(custom_fact.path)
+      stdout_str, status = Open3.capture2e(
+        "puppet device #{common_args} --deviceconfig #{device_conf.path} --pluginfactdest #{fact_dir} --apply 'if $facts[\"baz\"] != \"foo\" { fail(\"fact not found\") }'",
+      )
+      expect(stdout_str).not_to match %r{Error:}
```
DavidS: Please add a positive assertion, checking for a message iff the fact is there.

README.md
```diff
@@ -175,6 +175,48 @@ The provider needs to specify the `remote_resource` feature to enable the second
...
+```
+Note the call to return_custom_facts, which gets all the custom facts associated with that device.
+
+#### Custom facts for 'puppet device' - custom fact author
+
+Adding a custom fact only requires implemented the add_fact method, and putting it into the correct directory.
+This method gives access to the context/connection used by the device and the facts hash. 
+
+```
+class Example_ResourceAPI_Fact
```
DavidS: The example here is not inheriting from the `CustomFact` base class. I have no preference either way, but either the example should use the class, or the class should not be part of this PR.

README.md
```diff
@@ -175,6 +175,48 @@ The provider needs to specify the `remote_resource` feature to enable the second
...
+In your facts method:
+```
+def facts
+  facts = { 'foo' => 'bar' }
+  custom_facts = return_custom_facts
+  custom_facts.each do |custom_fact|
+    begin
+      load custom_fact
+      jim = Example_ResourceAPI_Fact
+      bla = jim.add_fact(nil, facts)
```
DavidS: should this be `self`, instead of `nil`?

README.md
```diff
@@ -175,6 +175,48 @@ The provider needs to specify the `remote_resource` feature to enable the second
...
+      end
+    end
+  facts
+end
+```
+Note the call to return_custom_facts, which gets all the custom facts associated with that device.
+
+#### Custom facts for 'puppet device' - custom fact author
+
+Adding a custom fact only requires implemented the add_fact method, and putting it into the correct directory.
```
timidri: @tphoney It's not yet clear to me what the step-by-step workflow is for the custom fact creator.

Questions:

1. The current PR for [cisco_ios](https://github.com/puppetlabs/cisco_ios/pull/234) seems to load facts from Example_ResourceAPI_Fact. This is obviously just an example, but what will be the general way of adding custom facts to Cisco IOS devices? Do I need to implement a custom provider + Device subclass for that, even if I just want to add a custom fact to an already supported device like cisco_ios?
2. Where and how do I deploy the custom fact class implementation?

PR #110, opened 2018-08-06
Allow an attribute with default boolean value to be set correctly

da-ar:


lib/puppet/resource_api.rb
```diff
@@ -178,9 +178,14 @@ def to_resource
             isnamevar
           end
 
+          type = Puppet::ResourceApi.parse_puppet_type(name, options[:type])
+
           # read-only values do not need type checking, but can have default values
-          if options[:behaviour] != :read_only
-            if options.key? :default
+          if options[:behaviour] != :read_only && options.key?(:default)
+            if type.is_a? Puppet::Pops::Types::PBooleanType
```
DavidS: What about compound types like `Optional[Boolean]`, `Array[Boolean]` and `Variant[String, Boolean]`?

lib/puppet/resource_api.rb
```diff
@@ -521,6 +524,15 @@ def self.mungify(type, value, error_msg_prefix)
       # the correct types, e.g. "1" (a string) to 1 (an integer)
       cleaned_value, error_msg = try_mungify(type, value, error_msg_prefix)
       raise Puppet::ResourceError, error_msg if error_msg
+    elsif type.is_a? Puppet::Pops::Types::PBooleanType
```
DavidS: See compound types above.

spec/puppet/resource_api_spec.rb
```diff
@@ -1034,6 +1052,74 @@ def set(_context, _changes); end
...
+      end
+
+      it { expect { described_class.register_type(definition) }.not_to raise_error }
+      it { expect(instance.flush).to eq(final_hash) }
+    end
+
+    context 'when the default value is false' do
+      let(:type) { Puppet::Type.type(:default_bool) }
+      let(:default_value) { false }
+      let(:instance) { type.new(name: 'foo', ensure: 'present') }
```
DavidS: lots of duplication here that can be pulled out to the parent context?

spec/puppet/resource_api_spec.rb
```diff
@@ -1344,6 +1430,7 @@ def set(_context, changes)
         context 'when flushing' do
           before(:each) do
             Puppet.debug = true
+            instance.my_provider.set(nil, nil) # reset the current_state
```
DavidS: Huh, how are the changes above leaking into this test case?

lib/puppet/resource_api.rb
DavidS: Would it be enough to apply the workaround if and only if the default value is actually `false`?

spec/puppet/resource_api_spec.rb
da-ar: believe random seed order was changed when the new test went in and exposed the bug

DavidS: ah, ok! then please put this in a separate commit with a short note about that, to inform future-us of what went down here.

DavidS:
Added some cleanups to the tests. After this has passed CI, we can merge and release?

PR #115, opened 2018-09-11
(PDK-1150) Allow providers to override :title when retrieving resources

da-ar:
This change facilitates composite namevar functionality.


seanmil:
This patch worked for me. I'm successfully returning and using a supplied `:title` with it. Thanks!

PR #116, opened 2018-09-11
(PDK-1150) Add support for returning a resource title from get

seanmil:
To better support composite namevars, allow the `get` method to return a `:title` key to set the resource's title.


DavidS:
See also #115 :-/

PR #128, opened 2018-09-25
Adjust announcement template to reality

DavidS:
* new examples
* new packages


DavidS:
the changes from today's announcement

clairecadman:
@DavidS not quite sure what you mean - do you mean that you can't use Puppet 6 packages or the resource_api module until the next agent release? If so, then I would say that, like you said, with the "next" release.

DavidS:
If folks are using puppet6 packages, they can't upgrade right now, but will have to wait until the next puppet6 release. There is the additional process complexity of not being certain what/when the next release will be.

clairecadman:
@DavidS How about adding to the end of that paragraph: "Note that if you are using Puppet 6 packages, you will have to wait until the next release to upgrade".

DavidS:
:+1:

PR #134, opened 2018-10-25
(PDK-1209) Fix the other call-sites of const_defined? and const_get

DavidS:
Same as #132, there are more const_get and const_defined? calls that
need to ignore inherited names.


DavidS:
The appveyor build will not come back today, since it is blocked behind three puppet builds, each of which is reportedly taking an hour+. I'll force-merge as soon as travis goes green.

PR #135, opened 2018-10-25
Release prep for v1.6.2

DavidS:


DavidS:
Overriding the lack of CI runs since this is a hotfix that was already tested once in #134 , and needs to go out before people upgrade to 1.6.1.

PR #140, opened 2018-11-26
Move parameter and property logic to separate classes

bpietraga:
In this proposal the PR the changes were extraction of parameter logic to:
- `Puppet::ResourceApi::Property`
- `Puppet::ResourceApi::ReadOnlyParameter`
- `Puppet::ResourceApi::Parameter`

and movement of value defaults and aliases logic to

- `Puppet::ResourceApi::ValueCreator`

Commits are in smaller chunks and can be changed / squashed.
@DavidS if you have any changes, suggestions please let me know. I can adjust the code.

### DIFF
--- a/lib/puppet/resource_api.rb
+++ b/lib/puppet/resource_api.rb
@@ -201,6 +201,12 @@ module Puppet::ResourceApi
             raise Puppet::DevError, "#{definition[:name]}.#{name} has no type"
           end
 
+          if options[:desc]
+            desc "#{options[:desc]} (a #{options[:type]})"
+          else
+            warn("#{definition[:name]}.#{name} has no docs")
+          end
+
           # The initialize takes passed hash with resource.
           # The method pass needed arguments, depending on parent to:
           # - Puppet::ResourceApi::Property
@@ -212,36 +218,31 @@ module Puppet::ResourceApi
             super(name, data_type, data_definition, resource_hash[:resource])
           end
 
-          if options[:desc]
-            desc "#{options[:desc]} (a #{options[:type]})"
-          else
-            warn("#{definition[:name]}.#{name} has no docs")
-          end
-
+          # get type class object for the parameter or property
           type = Puppet::ResourceApi::DataTypeHandling.parse_puppet_type(
             name,
             options[:type],
           )
-
-          # inside of parameter or property provide method to check and return
-          # data type.
-          define_method(:data_type) do
-            type
-          end
-
-          # inside of parameter or property provide method to check and return
-          # the definition hash object.
-          define_method(:data_definition) do
-            definition
-          end
 
           # provide hints to `puppet type generate` for better parsing
           if type.instance_of? Puppet::Pops::Types::POptionalType
             type = type.type
           end
 
-          # Initialize ValueCreator and call create_values which creates alias
-          # values and default values for properties and params.
+          # inside of parameter or property provide method to check and return
+          # data type
+          define_method(:data_type) do
+            type
+          end
+
+          # inside of parameter or property provide method to check and return
+          # the definition hash object
+          define_method(:data_definition) do
+            definition
+          end
+
+          # initialize ValueCreator and call create_values which makes alias
+          # values and default values for properties and params
           Puppet::ResourceApi::ValueCreator.new(
             self,
             type,
--- a/lib/puppet/resource_api.rb
+++ b/lib/puppet/resource_api.rb
@@ -6,6 +6,7 @@ require 'puppet/resource_api/property'
 require 'puppet/resource_api/puppet_context' unless RUBY_PLATFORM == 'java'
 require 'puppet/resource_api/read_only_parameter'
 require 'puppet/resource_api/type_definition'
+require 'puppet/resource_api/value_creator'
 require 'puppet/resource_api/version'
 require 'puppet/type'
 require 'puppet/util/network_device'
@@ -196,6 +197,10 @@ module Puppet::ResourceApi
         # the i.ex.
         # newparam(:param_name, parent: Puppet::ResourceApi::Parameter) do; end;
         send(param_or_property, name.to_sym, parent: parent) do
+          unless options[:type]
+            raise Puppet::DevError, "#{definition[:name]}.#{name} has no type"
+          end
+
           # The initialize takes passed hash with resource.
           # The method pass needed arguments, depending on parent to:
           # - Puppet::ResourceApi::Property
@@ -207,44 +212,10 @@ module Puppet::ResourceApi
             super(name, data_type, data_definition, resource_hash[:resource])
           end
 
-          unless options[:type]
-            raise Puppet::DevError, "#{definition[:name]}.#{name} has no type"
-          end
-
           if options[:desc]
             desc "#{options[:desc]} (a #{options[:type]})"
           else
             warn("#{definition[:name]}.#{name} has no docs")
-          end
-
-          if options[:behaviour] == :namevar
-            isnamevar
-          end
-
-          # read-only values do not need type checking, but can have default values
-          if options[:behaviour] != :read_only && options.key?(:default)
-            if options.key? :default
-              if options[:default] == false
-                # work around https://tickets.puppetlabs.com/browse/PUP-2368
-                defaultto :false # rubocop:disable Lint/BooleanSymbol
-              elsif options[:default] == true
-                # work around https://tickets.puppetlabs.com/browse/PUP-2368
-                defaultto :true # rubocop:disable Lint/BooleanSymbol
-              else
-                # marshal the default option to decouple that from the actual value.
-                # we cache the dumped value in `marshalled`, but use a block to unmarshal
-                # everytime the value is requested. Objects that can't be marshalled
-                # See https://stackoverflow.com/a/8206537/4918
-                marshalled = Marshal.dump(options[:default])
-                defaultto { Marshal.load(marshalled) } # rubocop:disable Security/MarshalLoad
-              end
-            end
-          end
-
-          if name == :ensure
-            def insync?(is)
-              rs_value.to_s == is.to_s
-            end
           end
 
           type = Puppet::ResourceApi::DataTypeHandling.parse_puppet_type(
@@ -264,44 +235,19 @@ module Puppet::ResourceApi
             definition
           end
 
-          # puppet symbolizes some values through puppet/parameter/value.rb (see .convert()), but (especially) Enums
-          # are strings. specifying a munge block here skips the value_collection fallback in puppet/parameter.rb's
-          # default .unsafe_munge() implementation.
-          munge { |v| v }
-
           # provide hints to `puppet type generate` for better parsing
           if type.instance_of? Puppet::Pops::Types::POptionalType
             type = type.type
           end
 
-          case type
-          when Puppet::Pops::Types::PStringType
-            # require any string value
-            Puppet::ResourceApi.def_newvalues(self, param_or_property, %r{})
-          when Puppet::Pops::Types::PBooleanType
-            Puppet::ResourceApi.def_newvalues(self, param_or_property, 'true', 'false')
-            aliasvalue true, 'true'
-            aliasvalue false, 'false'
-            aliasvalue :true, 'true' # rubocop:disable Lint/BooleanSymbol
-            aliasvalue :false, 'false' # rubocop:disable Lint/BooleanSymbol
-
-          when Puppet::Pops::Types::PIntegerType
-            Puppet::ResourceApi.def_newvalues(self, param_or_property, %r{^-?\d+$})
-          when Puppet::Pops::Types::PFloatType, Puppet::Pops::Types::PNumericType
-            Puppet::ResourceApi.def_newvalues(self, param_or_property, Puppet::Pops::Patterns::NUMERIC)
-          end
-
-          if param_or_property == :newproperty
-            # stop puppet from trying to call into the provider when
-            # no pre-defined values have been specified
-            # "This is not the provider you are looking for." -- Obi-Wan Kaniesobi.
-            def call_provider(value); end
-          end
-
-          case options[:type]
-          when 'Enum[present, absent]'
-            Puppet::ResourceApi.def_newvalues(self, param_or_property, 'absent', 'present')
-          end
+          # Initialize ValueCreator and call create_values which creates alias
+          # values and default values for properties and params.
+          Puppet::ResourceApi::ValueCreator.new(
+            self,
+            type,
+            param_or_property,
+            options,
+          ).create_values
         end
       end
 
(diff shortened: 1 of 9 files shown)


DavidS:
I'll have :eyes: on this later today.

lib/puppet/resource_api/parameter.rb
```diff
@@ -0,0 +1,48 @@
...
+module Puppet; module ResourceApi; end; end # predeclare the main module # rubocop:disable Style/Documentation,Style/ClassAndModuleChildren
+
+# Class containing parameter functionality for ResourceApi.
+class Puppet::ResourceApi::Parameter < Puppet::Parameter
+  # This initialize takes arguments and sets up new parameter.
+  # @param name the name used for reporting errors
+  # @param type the type instance
+  # @param definition the definition of the property
+  # @param resource the resource instance which is passed to the parent class.
+  def initialize(name, type, definiton, resource)
```
DavidS: To clean this up in relation to my comment above around the `data_type` function, we need to have a good look at how this interface looks and works to ensure that we do not just move from one maintenance head-ache to the other.

Looking at coupling and cohesion, the signature as it stands has problems:

* too high coupling to `resource_api.rb` through passing the full `definition`.
* not enough cohesion with core puppet by requiring unpacking and repacking of the `resource` parameter.
* using `type` instead of `data_type` - thanks to the overloading of terms, we **always** have to be very specific of which kind of type (data vs resource type) we mean
* lack of extensibility/future-proofing, by not passing in the `options` of this attribute

To address these issues, a better signature could look like this:

```ruby
def initialize(type_name, data_type, attribute_name, resource_hash)
```

* provides the immediately necessary data for the object with no additional processing required 
* lists parameters in an order similar to a declaration in typed languages (loosely modelled after something like `class Foo { int bar; }`, with `type_name` being `Foo`, `data_type` being `int`, and `attribute_name` being `bar` )
* re-use puppet core's `resource_hash`, which gets passed through to `super` unmodified

Looking back at `resource_api.rb`, usage would now look like this:

```ruby
define_method(:initialize) do |resource_hash|
  super(definition[:name], self.class.data_type, name, resource_hash)
```

PS: typo: `definiton` -> `definition`
PPS: it would be helpful to rename `name` to `attribute_name` in the entire loop in `resource_api.rb`.

lib/puppet/resource_api/value_creator.rb
```diff
@@ -0,0 +1,87 @@
...
+  # @param resource_class the class of selected resource to be extended
+  # @param data_type the type instance
+  # @param definition the definition of the property
+  # @param options the ResourceAPI options hash containing setup information for
+  # selected parameter or property
+  def initialize(resource_class, data_type, param_or_property, options = {})
+    @resource_class = resource_class
+    @data_type = data_type
+    @param_or_property = param_or_property
+    @options = options
```
DavidS: all this state here seems like a lot of work for little benefit. Instead you could make this a module, and  `create_values` (and all other functions) `module_function`s and pass in the required data directly:

```ruby
def create_values(resource_class, data_type, param_or_property, options = {})
  # ...
```

With it being a module, you'd probably could even include it into the Property/Parameter classes directly and completely get rid of passing through the `resource_class`.

lib/puppet/resource_api/value_creator.rb
```diff
@@ -0,0 +1,87 @@
...
+        # we cache the dumped value in `marshalled`, but use a block to
+        # unmarshal everytime the value is requested. Objects that can't be
+        # marshalled
+        # See https://stackoverflow.com/a/8206537/4918
+        marshalled = Marshal.dump(@options[:default])
+        @resource_class.defaultto { Marshal.load(marshalled) } # rubocop:disable Security/MarshalLoad
+      end
+    end
+
+    case @data_type
```
DavidS: This is the place where the `Optional` unpacking from https://github.com/puppetlabs/puppet-resource_api/blob/9dd87a93c8e0926a90745058df5005d311825426/lib/puppet/resource_api.rb#L299-L302 should go. `@data_type` must always be the full unmodified type from `parse_puppet_type` to avoid future confusion.

lib/puppet/resource_api.rb
```diff
@@ -191,144 +207,48 @@ def to_resource
...
-            Puppet::ResourceApi.def_newvalues(self, param_or_property, 'absent', 'present')
-          end
+          # initialize ValueCreator and call create_values which makes alias
+          # values and default values for properties and params
+          Puppet::ResourceApi::ValueCreator.new(
+            self,
+            type,
+            param_or_property,
+            options,
+          ).create_values
```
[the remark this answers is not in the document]
bpietraga: This sounds really good! Thank you!

lib/puppet/resource_api/parameter.rb
bpietraga: I will change it, it looks like much better choice that what it is now

lib/puppet/resource_api/value_creator.rb
bpietraga: Composition sounds like much better choice, thanks

bpietraga:
I've included all suggestions from the comments, excluding the movement to ValueCreator from `resource_api.rb` to Parameter and Property classes. I will do it tomorrow 🙂The old puppet specs are now ok,.

lib/puppet/resource_api/parameter.rb
```diff
@@ -0,0 +1,50 @@
...
+  def initialize(type_name, data_type, attribute_name, resource_hash)
+    @type_name = type_name
+    @data_type = data_type
+    @attribute_name = attribute_name
+    super(resource_hash) # Pass resource to parent Puppet class.
+  end
+
+  # This method handles return of assigned value from parameter.
+  # @return [value] the type value
+  def value # rubocop:disable Style/TrivialAccessors
```
DavidS: Missed that on the first round: the entire method can be replaced by 
```ruby
attr_reader :value
```

By convention as the first thing in the class.

bpietraga:
I wasn't sure about the `attr_reader` but this looks much cleaner. Thanks!
Also I've found that I've added unnecessary `munge` to `Puppet::ResourceApi::ReadOnlyParameter ` as its already inheriting it from `Puppet::ResourceApi::Parameter`.

Also I've noticed weird behaviour when calling the methods from `Puppet::ResourceApi::ValueCreator`.  `defaultto` and aliases are not working correctly when called from the `initialize` in resource. i.ex. `Puppet::ResourceApi::Parameter`. I'm afraid that debugging why the `newvalue` called on correct class is not working might take a while.

Current setup works ok with call to `Puppet::ResourceApi::ValueCreator` from the `Puppet::ResourceApi`.

If code is looks ok i would like move the calls to `ValueCreator` from the `Puppet::ResourceApi` to parameter and properties in next PR. I should be just different place to call the module.

If there is anything to change please write me 🙂

bpietraga:
In last commit I've moved the `call_to_provider(_value)` from `Puppet::ResourceApi::ValueCreator` method to `Puppet::ResourceApi::Property`.

bpietraga:
I've updated tests based on suggestions and also renamed commits to include (maint) to keep naming consistent. Thank you for the help in preparing this PR! 🙂

spec/puppet/resource_api/property_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+    context 'when name is :ensure' do
+      let(:attribute_name) { :ensure }
+
+      it 'the #insync? method is avaliable' do
+        expect(property.methods.include?(:insync?)).to eq true
+      end
+    end
+  end
+
+  describe '#should=(value)' do
```
DavidS: I think the `should`/`should=`/`value`/`rs_value` tests can be simplified by folding them all into a unified structure:

```
describe 'value munging and storage' do
  before(:each) do
    allow(Puppet::ResourceApi::DataTypeHandling).to receive(:mungify)
      .with(data_type, value, 'test_name.some_property', false)
      .and_return(munged_value)

    property.should = value
  end

  context 'when handling strings' do
    let(:value) { 'value' }
    let(:munged_value) { 'munged value' }

    it { expect(property.should).to eq 'munged value' }
    it { expect(property.rs_value).to eq 'munged value' }
    it { expect(property.value).to eq 'munged value' }
  end

  context 'when handling boolean true' do
    let(:value) { true }
    let(:munged_value) { true }
    let(:data_type) { Puppet::Pops::Types::PBooleanType.new }

    it { expect(property.should).to eq :true }
    it { expect(property.rs_value).to eq true }
    it { expect(property.value).to eq :true }
  end
end
```

That gets rid of poking into the class' private parts with `instance_variable_set`, and clearly defines the expectations around how the different interfaces (should/value/rs_value) return the stored value.

lib/puppet/resource_api/property.rb
```diff
@@ -0,0 +1,85 @@
...
+  # used internally
+  # @returns the final mungified value of this property
+  def rs_value
+    @should ? @should.first : @should
+  end
+
+  # method added only for the :ensure property, add option to check if the
+  # rs_value matches is
+  def def_insync?
+    self.class.send(:define_method, :insync?) { |is| rs_value.to_s == is.to_s }
```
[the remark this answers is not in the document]
bpietraga: Thank you for reply. This makes sense, i will add the additional class check 👍 

I've found one thing that the below test will fail even when the logic for adding `insync?` is removed. Looks like it is already defined in `Puppet::Property` which is parent of `Puppet::ResourceApi::Property`
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/property.rb#L309

This one fails, checked on Ruby `2.1.9`, Puppet `4`, and also the `2.5.1`, Puppet `5`
```
  subject(:property) do
    described_class.new(type_name, data_type, attribute_name, resource_hash)
  end

  let(:type_name) { 'test_name' }
  let(:attribute_name) { 'some_property' }
  let(:data_type) { Puppet::Pops::Types::PStringType.new(nil) }
  let(:resource_hash) { { resource: {} } }

  it 'has no #insync? method' do
    expect(property.methods).not_to include(:insync?)
  end
```

spec/puppet/resource_api/property_spec.rb
bpietraga: This proposed tests look much cleaner and nicer. 😄

spec/puppet/resource_api/read_only_parameter_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    context 'when value is not set' do
+      it { expect(read_only_parameter.rs_value).to eq nil }
+      it { expect(read_only_parameter.value).to eq nil }
+    end
+
+    context 'when value is already set' do
+      before(:each) do
+        read_only_parameter.instance_variable_set(:@value, 'value')
+      end
```
bpietraga: I've added test when the instance variable `@value` is set for read only parameter, I can remove it if it is not needed.

spec/puppet/resource_api/property_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+  describe 'should error handling' do
+    it 'calls mungify and reports its error' do
+      expect(Puppet::ResourceApi::DataTypeHandling).to receive(:mungify)
+        .and_raise Exception, 'error'
+
+      expect { property.should = 'value' }.to raise_error Exception, 'error'
+
+      expect(property.should).to eq nil
+    end
+  end
```
bpietraga: I've added test when the the `mungify` is returning error

lib/puppet/resource_api/property.rb
```diff
@@ -0,0 +1,86 @@
...
+  def rs_value
+    @should ? @should.first : @should
+  end
+
+  # method overloaded only for the :ensure property, add option to check if the
+  # rs_value matches is. Only if the class is child of
+  # Puppet::ResourceApi::Property.
+  def def_insync?
+    self.class.send(:define_method, :insync?) { |is| rs_value.to_s == is.to_s } unless self.class == Puppet::ResourceApi::Property
+  end
```
bpietraga: I've added information about overloading `insync?` method. The method is already in the `Puppet::Property` but with different behavior. Testing approach could be checking if class receives `define_method`.

`self.class.define_method(:insync?) { |is| rs_value.to_s == is.to_s }` won't work, as you cannot call directly `define_method`, it has to be trough `send`.

If you have different approach, I'm open for changing this. 🙂

bpietraga: The check is guard clause, this is rubocop suggestion.

bpietraga: Just commit name change

lib/puppet/resource_api/property.rb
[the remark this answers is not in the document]
DavidS: Apologies, `insync?` is also defined in the base `Property` class. Spelunking through the ruby docs I found [`Object.public_methods`](http://ruby-doc.org/core-2.1.9/Object.html#method-i-public_methods)

```
it 'has no #insync? method' do
  expect(property.public_methods(false)).not_to include(:insync?)
end
```

Running that also flagged up the `super` call as not able to access the other test case variables, so I changed the entire tests to 

```
  describe 'the special :ensure behaviour' do
    let(:ensure_property_class) do
      Class.new(described_class) do
        define_method(:initialize) do
          super('test_name',
            Puppet::Pops::Types::PEnumType.new(%w[absent present]),
            :ensure,
            { resource: {} })
        end
      end
    end
    let(:ensure_property) { ensure_property_class.new }

    it 'has a #insync? method' do
      expect(ensure_property.public_methods(false)).to include(:insync?)
    end

    describe '#insync?' do
      it 'compares using symbols'
    end
  end
```

and those are passing now - except for the `compares using symbols` that still needs to be implemented.

spec/puppet/resource_api/property_spec.rb
DavidS: There is also this `:ensure` handling that needs to be tested: https://github.com/puppetlabs/puppet-resource_api/blob/9dd87a93c8e0926a90745058df5005d311825426/lib/puppet/resource_api.rb#L232

spec/puppet/resource_api/read_only_parameter_spec.rb
DavidS: No, good call.

lib/puppet/resource_api/property.rb
DavidS: * See the other comment about using `public_methods(false)` instead of `methods` for the test
* call `define_method` directly, instead of using send

spec/puppet/resource_api/value_creator_spec.rb
```diff
@@ -0,0 +1,224 @@
...
+    end
+
+    context 'when resource class is property' do
+      it 'resource_class has #call_provider defined' do
+        expect(resource_class.method_defined?(:call_provider)).to eq(true)
+      end
+
+      context 'when property is :ensure' do
+        it 'calls #newvalue twice' do
+          expect(Puppet::ResourceApi::Property).to have_received(:newvalue).twice
```
DavidS: All these tests are very shallow. At least also test the received arguments using `with()` matchers (see https://relishapp.com/rspec/rspec-mocks/v/3-8/docs/setting-constraints/matching-arguments).

bpietraga: I've updated specs file :slightly_smiling_face:

PR #141, opened 2018-12-07
(FM-7597) RSAPI Transport register function and Schema validation

da-ar:


lib/puppet/resource_api/transport_schema_def.rb
```diff
@@ -0,0 +1,8 @@
+require 'puppet/resource_api/base_type_definition'
+# Provides accessor methods for the type being provided
+class Puppet::ResourceApi::TransportSchemaDef < Puppet::ResourceApi::BaseTypeDefinition
+  def initialize(definition)
+    raise Puppet::DevError, 'TransportSchemaDef requires definition to be a Hash' unless definition.is_a?(Hash)
+    super(definition, :connection_info)
+  end
+end
```
DavidS: Given how small `TransportSchemaDef` and `TypeDefinition` have become I'd pref to have them all in one file:

```
module Puppet::ResourceApi
  # pre-declare class
  class BaseTypeDefinition; end
 
  class Puppet::ResourceApi::TypeDefinition < Puppet::ResourceApi::BaseTypeDefinition
    # ...
  ...
  ..
  class BaseTypeDefinition
    # the rest
```

lib/puppet/resource_api/transport_schema_def.rb
```diff
@@ -0,0 +1,8 @@
+require 'puppet/resource_api/base_type_definition'
+# Provides accessor methods for the type being provided
+class Puppet::ResourceApi::TransportSchemaDef < Puppet::ResourceApi::BaseTypeDefinition
+  def initialize(definition)
+    raise Puppet::DevError, 'TransportSchemaDef requires definition to be a Hash' unless definition.is_a?(Hash)
```
DavidS: duplicate check - if you're doing this because of the class name in the error message, use `self.class.name` in the base class.

lib/puppet/resource_api/transport.rb
```diff
@@ -0,0 +1,20 @@
+require 'puppet/resource_api/transport_schema_def'
+# Remote target transport API
+module Puppet::ResourceApi::Transport
+  class << self
+    attr_reader :transports
```
DavidS: exposing the transports like this will allow folks to directly interact with the hash and add/modify/delete entries without validation. there is no need for the accessor yet.

PR #142, opened 2018-12-10
(maint) Validate Type Schema

da-ar:
This has been split out and updated from pr #141

### DIFF
--- a/lib/puppet/resource_api/type_definition.rb
+++ b/lib/puppet/resource_api/type_definition.rb
@@ -3,8 +3,8 @@ class Puppet::ResourceApi::TypeDefinition
   attr_reader :definition
 
   def initialize(definition)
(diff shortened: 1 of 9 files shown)


lib/puppet/resource_api/type_definition.rb
```diff
@@ -36,6 +38,24 @@ def feature?(feature)
     supported
   end
 
+  def validate_schema
+    return if attributes.nil?
+    raise Puppet::DevError, "Type definition must have a name" if name.nil?
```
DavidS: duplicates https://github.com/puppetlabs/puppet-resource_api/blob/3099283abf4f3fc1aa6b64788d926b22cebef0b0/lib/puppet/resource_api.rb#L22-L48

da-ar:
Still needs the test moved to their new home

PR #151, opened 2019-01-28
(FM-7690) Fix transports cache to be environment aware

da-ar:


lib/puppet/resource_api/transport.rb
```diff
@@ -37,17 +36,32 @@ def connect(name, connection_info)
...
+    lookup = Puppet.lookup(:current_environment) if Puppet.respond_to? :lookup
+    @environment =  if lookup.nil?
+                      :transports_default
+                    else
+                      lookup.name
+                    end
+    @transports ||= {}
+    @transports[@environment] ||= {}
   end
+  module_function :init_transports # rubocop:disable Style/AccessModifierDeclarations
```
DavidS: As this is not part of the official API, this should be marked `private`, not `module_function`. Now that I think about this, the `get_context` and `validate` above, too.

lib/puppet/resource_api/transport.rb
```diff
@@ -1,5 +1,4 @@
 # rubocop:disable Style/Documentation
-module Puppet; end
 module Puppet::ResourceApi; end
 # rubocop:enable Style/Documentation
```
DavidS: switch to end-of-line `rubocop:disable` comment to clean this up

PR #155, opened 2019-02-13
(PDK-1271) Allow a transport to be wrapped and used like a device

da-ar:


lib/puppet/resource_api/transport.rb
```diff
@@ -35,6 +35,12 @@ def connect(name, connection_info)
   end
   module_function :connect # rubocop:disable Style/AccessModifierDeclarations
 
+  def inject_device(name, transport)
+    transport_wrapper = Puppet::ResourceApi::Transport::Wrapper.new(name, transport)
+    Puppet::Util::NetworkDevice.instance_variable_set(:@current, transport_wrapper)
```
adreyer: I updated bolt to prefer `set_device` once that gets into puppet but I guess I forgot to push it up.

```
    if  Puppet::Util::NetworkDevice.respond_to?(:set_device)
      Puppet::Util::NetworkDevice.set_device(conn_info['type'], transport_wrapper)
    else
      Puppet::Util::NetworkDevice.instance_variable_set(:@current, transport_wrapper)
    end
```

da-ar: Yeah, this is exactly to code @DavidS suggested earlier today :)

lib/puppet/resource_api/transport.rb
```diff
@@ -28,13 +28,27 @@ def list
   module_function :list # rubocop:disable Style/AccessModifierDeclarations
 
   def connect(name, connection_info)
-    validate(name, connection_info)
+    # clone the connection_info parameter, as validate may modify the value being passed in
```
da-ar: @DavidS can you take a look at this, the intention is to address the issue that `validate` may modify the connection_info hash.

DavidS: fixing the validation to not modify the `connection_info` would reduce GC pressure and compute overhead.

da-ar: Removed the dup and updated the way validation works. Running tests it seems that leaving unknown keys/values in place has had no negative effects.

DavidS: :crossed_fingers:

PR #160, opened 2019-03-06
(FM-7829) Update README with transports examples

willmeek:


clairecadman:
@willmeek @da-ar I made a couple of minor edits. A couple of questions:

What is a "device shim"? 

Is the somewhere that explains the difference between a transport class and a device class? It is not super clear from this section, other than they are two different types of classes?

willmeek:
@clairecadman 

The 'device shim' is basically that small 'device' class. It is kept in for compatibility reasons - to make things continue to work as currently that is what puppet device uses.

'transport' is the brave new world way of doing things, in future land we will expect to find a transport class and call specific methods as called out here eg. transport.connect

willmeek:
TODO:
- Detail bolt keywords for transport schema

DavidS:
The changes to allowed keywords also needs to go into https://github.com/puppetlabs/puppet-specifications/blob/master/language/resource-api/README.md#transport

README.md
```diff
@@ -172,19 +172,136 @@ The `create`/`update`/`delete` methods get called by the `SimpleProvider` base-c
...
+An example of a corresponding schema:
+
+```ruby
+# lib/puppet/transport/device_type.rb
+Puppet::ResourceAPI.register_transport(
+  name: 'device_type', # points at class Puppet::Transport::DeviceType
+  desc: 'Connects to a device_type',
+  # features: [], # future extension points
+  connection_info: {
+    hostname: {
```
adreyer: We should update this to match the guidelines below with host and user.

README.md
```diff
@@ -172,19 +172,136 @@ The `create`/`update`/`delete` methods get called by the `SimpleProvider` base-c
...
+* `remote-*` Any key starting with `remote-` is reserved for future use.
+ 
+Note: Currently bolt inventory requires that a name be set for every target and always uses that name as the URI. This means there is no way to specify `host` separately from the host section of the `name` when parsed as a URI.
+
+After the device class, transport class and transport schema have been implemented, `puppet device` will be able to use the new provider, and supply it (through the device class) with the URL specified in the [`device.conf`](https://puppet.com/docs/puppet/5.3/config_file_device.html).
+
+#### Transport/device specific providers
+
+To allow modules to deal with different backends independently, the Resource API implements a mechanism to use different API providers side by side. For a given transport/device class (see above), the Resource API will first try to load a `Puppet::Provider::TypeName::DeviceType` class from `lib/puppet/provider/type_name/device_type.rb`, before falling back to the regular provider at `Puppet::Provider::TypeName::TypeName`.
```
adreyer: I think `Puppet::Provider::TypeName::<DeviceType>` is clearer here.

README.md
```diff
@@ -172,19 +172,136 @@ The `create`/`update`/`delete` methods get called by the `SimpleProvider` base-c
...
+* `protocol` - Use to specify which protocol the transport should use for example `http`, `https`, `ssh` or `tcp`
+* `user` - The user the transport should connect as.
+* `port` - The port the transport should connect to.
+ 
+###### Non-Usable within the schema
+ 
+The following keywords are keywords that must not be used by the transport schema:
+ 
+* `name` - transports should use `uri` instead of name.
+* `path`
```
adreyer: both path and query can be documented as "reserved as a uri part".

PR #164, opened 2019-03-14
(MAINT) Bump version

sebastian-miclea:


DavidS:
This fails because
```
Gem::RuntimeRequirementNotMetError: parallel requires Ruby version >= 2.2. The
current ruby version is 2.1.0.
An error occurred while installing parallel (1.14.0), and Bundler
cannot continue.
Make sure that `gem install parallel -v '1.14.0' --source
'https://rubygems.org/'` succeeds before bundling.

In Gemfile:
  rubocop-rspec was resolved to 1.27.0, which depends on
    rubocop was resolved to 0.57.2, which depends on
      parallel
```

Interestingly enough this passed for the puppet4 gem (using parallel 1.13.0), but failed for the 4.10.x git branch. I've re-kicked the puppet4 gem cell to make sure there were no shenaningans with rubygems.org going on, but it is still passing.

DavidS:
Updated this PR with the Gemfile fixes from #166 (which probably will become obsolete through the mergup), and fixed the version bump to be in the right commit.

PR #168, opened 2019-04-10
(FM-7839) Implement `to_json` method for ResourceShim

da-ar:
bolt's `get_resources` uses the `libexec/query_resources.rb` to retrieve
resources as json from the target nodes. Especially https://github.com/puppetlabs/bolt/blob/7dec6e49d66c7f9ae06088d46ee3330de0ed2fed/libexec/query_resources.rb#L71
causes `ResourceShims` to fail as they did not implement `to_json`.

This integration test models what `libexec/query_resources.rb` is doing.
The added `to_json(*)` method now passes this test, and - shown in manual
testing - works for bolt.

### DIFF
--- a/lib/puppet/resource_api/glue.rb
+++ b/lib/puppet/resource_api/glue.rb
@@ -41,6 +41,13 @@ module Puppet::ResourceApi
       YAML.dump('type' => { title => attributes }).split("\n").drop(2).join("\n") + "\n"
     end
 
+    def to_json(*)
+      attrs = filtered_keys.map { |k| [k.to_s, values[k]] unless values[k].nil? }
+      attributes = Hash[*attrs.compact.flatten]
+      resource = { title => attributes }
+      resource.to_json
+    end
+
     # attribute names that are not title or namevars
     def filtered_keys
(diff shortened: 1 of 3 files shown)


lib/puppet/resource_api/glue.rb
```diff
@@ -41,6 +41,13 @@ def to_hierayaml
       YAML.dump('type' => { title => attributes }).split("\n").drop(2).join("\n") + "\n"
     end
 
+    def to_json
```
adreyer: Is there something we include or a base class we could use that would make this just work?

DavidS:
After a bit more digging I could identify the cause of @adreyer 's issues in https://github.com/puppetlabs/bolt/blob/7dec6e49d66c7f9ae06088d46ee3330de0ed2fed/libexec/query_resources.rb#L71 . This integration test and `to_json` method will fix the issue with bolt.

:shipit:

DavidS:
closing and re-opening to tickle travis

lib/puppet/resource_api/glue.rb
DavidS: I haven't found where `to_json` is originally defined, and the whole point of the `glue.rb` is to avoid the secret handshakes required to initialise puppet's `Resource` structs, and also to provide a living documentation of the _actual_ API required on those objects.

PR #174, opened 2019-05-21
(MODULES-9428) make the composite namevar implementation usable

DavidS:
This is a series of changes to make the composite namevar behaviour work. See the individual commits for details and https://github.com/puppetlabs/puppet-specifications/pull/140 for the specification changes/clarifications.

PR #175, opened 2019-05-21
(maint) make test order really random

DavidS:
The test leakage fixed in #174 was only detected by
changing the test order. Removing the seed will ensure that different
test orderings are tested, leading to a more robust test suite as new
issues are discovered and fixed.

This already has found another issue fixed here too.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,3 +1,3 @@
 --format documentation
 --color
---order rand:123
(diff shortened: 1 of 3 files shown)


DavidS:
Please hold off merging this until #186 has landed.

PR #177, opened 2019-05-30
(maint) implement `desc`/`docs` fallback

DavidS:
This implements the changes specified in https://github.com/puppetlabs/puppet-specifications/pull/141

> The text and examples have been inconsistent with how `desc` vs `docs`
> has been handled. This change fixes the text and examples to all show
> and require `desc`, but accept `docs` in places where we showed it
> previously.


DavidS:
The failing tests also fail on `1.6.x` HEAD: https://travis-ci.org/puppetlabs/puppet-resource_api/builds/539161298

We can progress with this change and ticket up the puppet problem.

DavidS:
Including build fix from #184 .

DavidS:
Rebasing for clean history after the tests preliminary passed.

PR #180, opened 2019-06-10
(maint) Add to_hash function to resourceShim for compatibility

da-ar:


spec/puppet/resource_api/glue_spec.rb
```diff
@@ -56,5 +56,9 @@
         it { expect(instance.to_hierayaml).to eq "  ? |-\n    foo:\n    bar\n  : attr: value\n    attr_ro: fixed\n" }
       end
     end
+
+    describe '.to_hash' do
+      it { expect(instance.values).to eq(namevarname: 'title', attr: 'value', attr_ro: 'fixed') }
```
DavidS: `.values` should be `to_hash` right?

da-ar: 🤦‍♂

PR #182, opened 2019-06-11
(PUP-9747) Relax validation for bolt

DavidS:
This is an initial draft of the solution. Still needs tests and validation.

### DIFF
--- a/lib/puppet/resource_api/transport.rb
+++ b/lib/puppet/resource_api/transport.rb
@@ -53,6 +53,11 @@ module Puppet::ResourceApi::Transport
         environment: current_environment,
       }
     end
+
+    if connection_info.key?(:"remote-transport")
+      clean_bolt_attributes(transport_schema, connection_info)
+    end
+
     message_prefix = 'The connection info provided does not match the Transport Schema'
     transport_schema.check_schema(connection_info, message_prefix)
     transport_schema.validate(connection_info)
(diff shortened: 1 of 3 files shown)


lib/puppet/resource_api/transport.rb
```diff
@@ -53,6 +53,31 @@ def self.validate(name, connection_info)
         environment: current_environment,
       }
     end
+
+    # Attributes we expect from bolt, but want to ignore if the transport does not expect them
+    [:uri, :host, :protocol, :user, :port].each do |attribute_name|
```
adreyer: I think we need to add password to this list.

lib/puppet/resource_api/transport.rb
```diff
@@ -53,6 +53,31 @@ def self.validate(name, connection_info)
...
+      if connection_info.key?(attribute_name) && !transports[name].attributes.key?(attribute_name)
+        discard_known_good_attribute(transport_name, attribute_name)
+        connection_info.delete(attribute_name)
+      end
+    end
+
+    # Attributes that bolt emits, but we want to ignore if the transport does not expect them
+    ([:name, :path, :query, :"run-on", :"remote-transport", :implementations] + connection_info.keys.grep { |k| k.to_s.starts_with? 'remote-' }).each do |attribute_name|
```
adreyer: for future-proofness I think we need to add `remote-*` to this list.

it's unclear to me how `implementations` ends up here but I don't see any harm in including it.

DavidS:
This still needs to be validated with the bolt tests. I'll get on that next.

adreyer:
I have gotten bolt's test passing with this branch.

PR #185, opened 2019-06-13
(maint) test fixes

DavidS:
These test fixes are fixes for unclean tests uncovered by `--rand` test ordering.


DavidS:
rebased on top of current master

PR #200, opened 2019-08-28
(FM-8092) Fix caching scope of transport schemas

DavidS:
When running in puppetserver for the [environment_transports](https://github.com/puppetlabs/puppetserver/blob/master/documentation/puppet-api/v3/environment_transports.markdown)
API, the Resource API would cache registered transport schemas forever.
This meant that deploying a new module version and running the usual
environment cache cleaning scripts would not update the transport schema.

With this change, we push the transport cache into an ObjectIdCacheAdapter
around the environment, which will be cleaned up by the internal utilities.

### DIFF
--- a/spec/puppet/resource_api/transport_spec.rb
+++ b/spec/puppet/resource_api/transport_spec.rb
@@ -5,9 +5,9 @@ RSpec.describe Puppet::ResourceApi::Transport do
     environment = class_double(Puppet::Node::Environment)
 
     if name.nil?
-      allow(Puppet).to receive(:respond_to?).and_return(false)
+      allow(Puppet).to receive(:respond_to?).with(:lookup).and_return(false)
     else
-      allow(Puppet).to receive(:respond_to?).and_return(true)
+      allow(Puppet).to receive(:respond_to?).with(:lookup).and_return(true)
     end
 
     allow(Puppet).to receive(:lookup).with(:current_environment).and_return(environment)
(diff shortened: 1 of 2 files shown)


hlindberg:
Looks ok to me - (I assume it writes to the  cache by getting the `transports` hash - which happens to be the hash managed by the cache adapter).

spec/puppet/resource_api/transport_spec.rb
```diff
@@ -73,64 +73,16 @@ def change_environment(name = nil)
...
-          expect(transports['foo'][schema2[:name]]).to be_a_kind_of(Puppet::ResourceApi::TransportSchemaDef)
-          expect(transports['foo'][schema2[:name]].definition).to eq(schema2)
-
-          change_environment(:bar)
-          described_class.register(schema3)
-          expect(transports).to have_key(:bar)
-          expect(transports[:bar][schema3[:name]]).to be_a_kind_of(Puppet::ResourceApi::TransportSchemaDef)
-          expect(transports[:bar][schema3[:name]].definition).to eq(schema3)
+        it 'adds to the transports register' do
+          expect { described_class.register(schema) }.not_to raise_error
```
da-ar: missing tests that handle caches other than `transports_default`

DavidS: added a environment to the other test.

PR #236, opened 2019-10-02
(packaging) Bump to version '1.9.0' [no-promote]

GabrielNagy:
The automatic bump script missed this since version 1.9.0 was previously committed/reverted in 1edc1b583655393d747ea0ac48f1718ddc4d8889

### DIFF
--- a/lib/puppet/resource_api/version.rb
+++ b/lib/puppet/resource_api/version.rb
@@ -1,5 +1,5 @@
 module Puppet
   module ResourceApi
-    VERSION = '1.8.8'.freeze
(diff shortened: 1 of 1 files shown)


DavidS:
@mihaibuzgau how does this tie up with the plans for lovejoy branching? we still have a bug or two in-flight that might need to get promoted into lovejoy. are they supposed to get shipped as 1.9.x or will there be a 1.8.x branch in time for this to happen?

Otherwise I'd rather hold off on this version bump until lovejoy is out the door to avoid additional confusion.

PR #238, opened 2019-10-03
(FM-8336) Capture and expose attribute ordering from transport schema

DavidS:
When exposing attributes in the UI, they currently need to be alphabetised by their name, since the transmission through APIs does not preserve key order of the connection_info hash.

To facilitate showing attributes in the order defined by the module author, this change embeds the key-insertion order from the TypeDefinition into the API response.


DavidS:
This supersedes https://github.com/puppetlabs/puppetserver/pull/2193

Issue #214, opened 2019-09-24
failing resource poisons the context/flush implementation


DavidS:
## Describe the Bug
When puppet applies two resources of the same type, and the first one fails, but the second one would succeed, the lack of a reset of the context.failed? status in resource_api.rb:388 causes the second resource to fail too.

## Expected Behavior
The second resource should be evaluated independently of the first resource's success.

## Additional Context
Originally filed as https://tickets.puppetlabs.com/browse/PDK-1134

PR #241, opened 2019-10-28
Reset context.failed? between resources

seanmil:
Ensure that context.failed? is reset after a failure encountered
during a flush so following resources do not falsely indicate
they encountered an error.


seanmil:
This should address issue #214

PR #253, opened 2020-02-01
add Ruby 2.6.5 to test matrix

bastelfreak:
Ruby 2.6 was released over a year ago so I think it makes sense to test against it.


DavidS:
Hi @bastelfreak , the resource api is only testing on ruby versions that we actually ship, hence I'm closing this. We'll add more ruby versions when we get ready to ship a newer ruby in the agent packaging.

PR #254, opened 2020-02-01
Add Ruby 2.7.0 to Travis-CI matrix

bastelfreak:
Ruby 2.7 was released some time ago and the first distributions start to
ship it. Based on that I think it's a good idea to also run tests
against it.


bastelfreak:
Depends on https://github.com/puppetlabs/puppet/pull/7919

DavidS:
Hi @bastelfreak , the resource api is only testing on ruby versions that we actually ship, hence I'm closing this. We'll add more ruby versions when we get ready to ship a newer ruby in the agent packaging.

PR #255, opened 2020-02-01
Add `puppet` as runtime dependency

bastelfreak:
several classes from the puppet gem are loaded within this lib, so it's
a runtime dependency.

This only specifies the absolute lowest bound that is required to make
the code load. Actual requirements, bug-for-bug compatibility and
commercial support options are a lot more complex than can be expressed
here.

Signed-off-by: David Schmitt <david.schmitt@puppet.com>

### DIFF
--- a/puppet-resource_api.gemspec
+++ b/puppet-resource_api.gemspec
@@ -26,4 +26,5 @@ Gem::Specification.new do |spec|
   spec.require_paths = ['lib']
 
   spec.add_runtime_dependency 'hocon', '>= 1.0'
+  spec.add_runtime_dependency 'puppet', '>= 4.0'
 end


bastelfreak:
I'm not sure if version 6 is the correct requirement. Would 5 work as well?

DavidS:
this should work from puppet 4.10; since that v is not supported anymore, 5.5 should be the lower bound.

bastelfreak:
mhm, this requires some rework on the travis configuration

daianamezdrea:
Hello @bastelfreak, this is WIP ? Thank you!

DavidS:
Rebased on top of #275 to include the other dependency fixes.

DavidS:
bundler 1.17 is very unhappy over this change in conjunction with this library being a part of puppet 6:
```
Resolving dependencies...

Your bundle requires gems that depend on each other, creating an infinite loop.

Please remove either gem 'puppet' or gem 'puppet-resource_api' and try again.
```
-- from https://travis-ci.org/github/puppetlabs/puppet-resource_api/jobs/696802683

PR #263, opened 2020-03-20
(maint) Mock Hocon.load(...)

IrimieBogdan:
Mock Hocon.load(...) as it is used by Facter 4 before it is used in the class under test.


DavidS:
Interesting, master started to pass again last week (https://travis-ci.org/puppetlabs/puppet-resource_api/builds/668120415)  using facter 4.0.13 after failing earlier with facter 4.0.12. Is this still needed?

IrimieBogdan:
The tests from `Puppet::ResourceApi::Transport::Wrapper` call Puppet, which in turn calls `Facter.debug(....)`. `Facter.debug()` can read the facter config file `facter.conf` (if the file exists) using `Hocon.load(...)`. If this happens, the test will fail, because of `Hocon.load(...)` mock. To safeguard agains this edge case, we should consider this PR.

DavidS:
Thanks for the explanation.

PR #284, opened 2021-03-30
Enable catalog access feature

kreeuwijk:
This PR enables setting a `raw_catalog_access` feature flag on the type, which then provides access to the catalog in the provider by using `context.catalog`.

### DIFF
--- a/lib/puppet/resource_api.rb
+++ b/lib/puppet/resource_api.rb
@@ -430,7 +430,7 @@ MESSAGE
       end
 
       def context
-        @context ||= PuppetContext.new(TypeDefinition.new(definition), type_definition.feature?('raw_catalog_access') ? @catalog : nil)
+        @context ||= PuppetContext.new(type_definition, type_definition.feature?('raw_catalog_access') ? @catalog : nil)
       end
 
       def self.title_patterns
--- a/lib/puppet/resource_api.rb
+++ b/lib/puppet/resource_api.rb
@@ -430,8 +430,7 @@ MESSAGE
       end
 
       def context
-        self.class.context.catalog = @catalog if type_definition.feature?('raw_catalog_access')
-        self.class.context
+        @context ||= PuppetContext.new(TypeDefinition.new(definition), type_definition.feature?('raw_catalog_access') ? @catalog : nil)
       end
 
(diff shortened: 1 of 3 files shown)


lib/puppet/resource_api.rb
```diff
@@ -430,6 +430,7 @@ def strict_check_title_parameter(current_state)
       end
 
       def context
+        self.class.context.catalog = @catalog if type_definition.feature?('raw_catalog_access')
         self.class.context
       end
```
DavidS, line 435:
```suggestion
      def context
        @context ||= PuppetContext.new(TypeDefinition.new(definition), type_definition.feature?('raw_catalog_access') ? @catalog : nil)
      end
```

By creating the context for each instance separately, we can avoid the write access to `context.catalog` (see below).

DavidS (changes requested):
On preliminary testing, the class-level context doesn't have access to the catalog, but is used during `instances` for calling `get`. This makes me think that I'm not fully understanding the object lifecycle around this, making me afraid of either running into issues where the catalog is not yet available (as the unit tests flag up) or the catalog leaks through the agent process across multiple invocations.

This is also limited in that the catalog is not available during `puppet resource` calls. While providers need to be prepared to handle this (and we need to call it out in docs), I don't think there's any way around this, as `puppet resource` just doesn't have a catalog.

Overall this will need a bit more investigation/exploration/change before we can merge it.

michaeltlombardi:
@kreeuwijk I did a minor fixup to get existing tests passing and manually validated that things are working, but I think we'll need some more extensive testing (acceptance and unit alike) and noodling out to be sure there's no weird side effects or changing during a run.

DavidS:
Testing this might be a bit tricky as this will require a running puppet agent service and then verifying that the catalog from the first run doesn't leak into the subsequent runs. Not really sure how to set this up properly.

jpogran:
With the inability to add automated tests at the moment, the risk of adding a new API surface that exposes an advanced raw access to the catalog which is contrary to the purpose of this simpler API, and the uncertainty of the clear purpose/need for the feature I am closing this PR. Thank you for your contribution, if you feel this should be included please reach out to myself and @binford2k

Issue #225, opened 2019-09-24
Allow custom insync? checking for resources


DavidS:
## Use Case
For some use-cases - like specifying upper/lower bounds on set memberships ("for instance, I have a use-case where I want things to be treated as in-sync as long as the array specified in Puppet is a equal to or a subset of the array in the response from the API") - the Resource API type schema is not expressive enough. This leads to nasty workarounds, like munging up the desired values in canonicalize.

The current restriction is so that anyone dealing with raw resource data doesn't need to run the provider code to understand whether something is insync or not. 

## Describe the Solution You Would Like
???

## Describe Alternatives You've Considered
Add a `comparison_operator` key (with values like `eq`, `min`, `max`) to the attribute definition.

See ASN.1/LDAP schemas for an example of data-defined comparisons.

## Additional Context
Originally filed as https://tickets.puppetlabs.com/browse/MODULES-9574.

seanmil:
Not having a way of defining an `insync?` / `compare` method has been causing me a lot of headache recently and - so far - I have resorted to the nasty munging in `canonicalize` method. I recalled seeing a ticket on this so I came here to determine if it was something I could implement and the method I was considering implementing seems not to be in favor (adding a `compare` type function).

Has any more thought been given to how else this might be implemented? A `comparison_operator` key might be a good start for "slightly complicated" comparison cases, but I still feel that having something more flexible is ultimately required.

DavidS:
@seanmil With the restrictions specified in the description, I'm currently having no good idea on how to implement that (which usually means I'll have to compromise on the reqs eventually). Providing a list of specific usecases would definitely help either way (finding better ideas and putting pressure on the requirements).

seanmil:
The current use case I have in mind is something like the `user` type's `inclusive` vs. `minimum` option which affects how the group state is determined.

This is not (as far as I can tell) possible to currently achieve, even via abuse of canonicalize, but would be simple to achieve if a `compare()` method were available.

It might be possible with a sophisticated enough ASN.1/LDAP style filter/query syntax, but I'm not sure that complexity of designing and building something like that outweighs having a `compare()` function available and, even with some type of LDAP-style data-defined comparison I suspect there would be limits on what you could express, which would still leave some folks needing a `compare()` function.

Here is what I propose: add support for a `compare(is, should)` method that returns a true if a resource needs to be synchronized, or a false if they do not. Enable it with a feature flag, e.g. `custom_compare`.

[...]

PR #285, opened 2021-04-09
(GH-225) Add support for custom insync

michaeltlombardi:
This PR defines a custom `insync?` method for properties of a Resource API Type; this method can *only* be added to properties, not parameters, and is only called for properties which are not namevars.

Currently, it does a few things:

1. Add `custom_insync` to the list of valid type features
2. Ensure that if the `custom_insync` feature is specified all properties **except** the `ensure` property get the custom `insync?`, which calls into the provider before supering back to the property `insync?` method as defined by Puppet. `ensure` is already using a custom `insync?`, which this PR leaves alone.
3. Passes (some) information about the type and current context to the property so that the provider is available to be called from the new `insync?` method and context can be accessed from that method in the provider.

What this currently enables:

- A developer can now specify the `custom_insync` feature for a type and an `insync?` method in their provider; that method can know the attribute being checked, the `is` and `should` values of the *resource*, & the current context. If nothing is specified in this method, the comparison defaults to what Puppet currently uses; the same is true for any properties not especially handled inside of this method. Any comparison that returns `nil` will be supered back to the existing `insync?` method defined in `Puppet::Property`.

### DIFF
--- a/lib/puppet/resource_api.rb
+++ b/lib/puppet/resource_api.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'pathname'
 require 'puppet/resource_api/data_type_handling'
 require 'puppet/resource_api/glue'
--- a/lib/puppet/resource_api.rb
+++ b/lib/puppet/resource_api.rb
@@ -123,6 +123,22 @@ module Puppet::ResourceApi
         @rsapi_title
       end
 
+      def rsapi_canonicalized_target_state
+        @rsapi_canonicalized_target_state ||= begin
+          # skip puppet's injected metaparams
+          actual_params = @parameters.select { |k, _v| type_definition.attributes.key? k }
+          target_state = Hash[actual_params.map { |k, v| [k, v.rs_value] }]
+          target_state = my_provider.canonicalize(context, [target_state]).first if type_definition.feature?('canonicalize')
+          target_state
+        end
+        @rsapi_canonicalized_target_state
+      end
+
+      def rsapi_current_state
+        refresh_current_state unless @rsapi_current_state
+        @rsapi_current_state
+      end
+
       def to_resource
         to_resource_shim(super)
       end
@@ -166,6 +182,21 @@ module Puppet::ResourceApi
         raise_missing_params if @missing_params.any?
       end
 
+      # If the custom_insync feature is specified but no insyncable attributes are included
+      # in the definition, add the hidden rsapi_custom_insync_trigger property.
+      # This property exists *only* to allow a resource without properties to still execute an
+      # insync check; there's no point in specifying it in a manifest as it can only have one
+      # value; it cannot be specified in a type definition as it should only exist in this case.
+      if type_definition.feature?('custom_insync') && type_definition.insyncable_attributes.empty?
+        custom_insync_trigger_options = {
+          type: 'Enum[do_not_specify_in_manifest]',
+          desc: 'A hidden property which enables a type with custom insync to perform an insync check without specifying any insyncable properties',
+          default: 'do_not_specify_in_manifest',
+        }
+
+        type_definition.create_attribute_in(self, :rsapi_custom_insync_trigger, :newproperty, Puppet::ResourceApi::Property, custom_insync_trigger_options)
+      end
+
       definition[:attributes].each do |name, options|
         # puts "#{name}: #{options.inspect}"
 
@@ -189,43 +220,7 @@ module Puppet::ResourceApi
           parent = Puppet::ResourceApi::Property
         end
 
-        # This call creates a new parameter or property with all work-arounds or
-        # customizations required by the Resource API applied. Under the hood,
-        # this maps to the relevant DSL methods in Puppet::Type. See
-        # https://puppet.com/docs/puppet/6.0/custom_types.html#reference-5883
-        # for details.
-        send(param_or_property, name.to_sym, parent: parent) do
-          if options[:desc]
-            desc "#{options[:desc]} (a #{options[:type]})"
-          end
-
-          # The initialize method is called when puppet core starts building up
-          # type objects. The core passes in a hash of shape { resource:
-          # #<Puppet::Type::TypeName> }. We use this to pass through the
-          # required configuration data to the parent (see
-          # Puppet::ResourceApi::Property, Puppet::ResourceApi::Parameter and
-          # Puppet::ResourceApi::ReadOnlyParameter).
-          define_method(:initialize) do |resource_hash|
-            super(definition[:name], self.class.data_type, name, resource_hash)
-          end
-
-          # get pops data type object for this parameter or property
-          define_singleton_method(:data_type) do
-            @rsapi_data_type ||= Puppet::ResourceApi::DataTypeHandling.parse_puppet_type(
-              name,
-              options[:type],
-            )
-          end
-
-          # from ValueCreator call create_values which makes alias values and
-          # default values for properties and params
-          Puppet::ResourceApi::ValueCreator.create_values(
-            self,
-            data_type,
-            param_or_property,
-            options,
-          )
-        end
+        type_definition.create_attribute_in(self, name, param_or_property, parent, options)
       end
 
       def self.instances
@@ -279,11 +274,9 @@ module Puppet::ResourceApi
       end
 
       def retrieve
-        refresh_current_state unless @rsapi_current_state
+        Puppet.debug("Current State: #{rsapi_current_state.inspect}")
 
-        Puppet.debug("Current State: #{@rsapi_current_state.inspect}")
-
-        result = Puppet::Resource.new(self.class, title, parameters: @rsapi_current_state)
+        result = Puppet::Resource.new(self.class, title, parameters: rsapi_current_state)
         # puppet needs ensure to be a symbol
         result[:ensure] = result[:ensure].to_sym if type_definition.ensurable? && result[:ensure].is_a?(String)
 
@@ -302,10 +295,7 @@ module Puppet::ResourceApi
         raise_missing_attrs
 
         # puts 'flush'
-        # skip puppet's injected metaparams
-        actual_params = @parameters.select { |k, _v| type_definition.attributes.key? k }
-        target_state = Hash[actual_params.map { |k, v| [k, v.rs_value] }]
-        target_state = my_provider.canonicalize(context, [target_state]).first if type_definition.feature?('canonicalize')
+        target_state = rsapi_canonicalized_target_state
 
         retrieve unless @rsapi_current_state
 
--- a/lib/puppet/resource_api/parameter.rb
+++ b/lib/puppet/resource_api/parameter.rb
@@ -13,7 +13,7 @@ class Puppet::ResourceApi::Parameter < Puppet::Parameter
   # @param attribute_name the name of attribue of the parameter
   # @param resource_hash the resource hash instance which is passed to the
   # parent class.
-  def initialize(type_name, data_type, attribute_name, resource_hash)
+  def initialize(type_name, data_type, attribute_name, resource_hash, _referrable_type = nil)
     @type_name = type_name
     @data_type = data_type
     @attribute_name = attribute_name
--- a/lib/puppet/resource_api/parameter.rb
+++ b/lib/puppet/resource_api/parameter.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'puppet/util'
 require 'puppet/parameter'
 
--- a/lib/puppet/resource_api/property.rb
+++ b/lib/puppet/resource_api/property.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'puppet/util'
 require 'puppet/property'
 
--- a/lib/puppet/resource_api/property.rb
+++ b/lib/puppet/resource_api/property.rb
@@ -102,7 +102,11 @@ class Puppet::ResourceApi::Property < Puppet::Property
 
       raise(Puppet::DevError, 'No insync? method defined in the provider; an insync? method must be defined if the custom_insync feature is defined for the type') unless provider.respond_to?(:insync?)
 
-      provider_insync_result = provider.insync?(context, title, @attribute_name, is_hash, should_hash)
+      provider_insync_result, change_message = provider.insync?(context, title, @attribute_name, is_hash, should_hash)
+
+      unless provider_insync_result.nil? || change_message.nil? || change_message.empty?
+        @change_to_s_value = change_message
+      end
 
       case provider_insync_result
       when nil
@@ -111,16 +115,10 @@ class Puppet::ResourceApi::Property < Puppet::Property
         # Otherwise, super and rely on Puppet::Property.insync?
         super(is)
       when TrueClass, FalseClass
-        # When returning an actual boolean, use default change messaging
         return provider_insync_result
-      when String
-        # When returning a string, set the change message to the result and return false
-        @change_to_s_value = provider_insync_result
-        return false
       else
-        # When returning anything else, warn about a non-idiomatic return and return false
-        context.warning("Custom insync for #{@attribute_name} returned a #{provider_insync_result.class} instead of a String")
-        return false
+        # When returning anything else, raise a DevError for a non-idiomatic return
+        raise(Puppet::DevError, "Custom insync for #{@attribute_name} returned a #{provider_insync_result.class} with a value of #{provider_insync_result.inspect} instead of true/false; insync? MUST return nil or the boolean true or false") # rubocop:disable Metrics/LineLength
       end
     end
 
--- a/lib/puppet/resource_api/property.rb
+++ b/lib/puppet/resource_api/property.rb
@@ -11,12 +11,29 @@ class Puppet::ResourceApi::Property < Puppet::Property
   # @param attribute_name the name of attribue of the property
   # @param resource_hash the resource hash instance which is passed to the
   # parent class.
-  def initialize(type_name, data_type, attribute_name, resource_hash)
+  def initialize(type_name, data_type, attribute_name, resource_hash, referrable_type = nil)
     @type_name = type_name
     @data_type = data_type
     @attribute_name = attribute_name
-    # Define class method insync?(is) if the name is :ensure
-    def_insync? if @attribute_name == :ensure && self.class != Puppet::ResourceApi::Property
+    @resource = resource_hash[:resource]
+    @referrable_type = referrable_type
+
+    # Do not want to define insync on the base class because
+    # this overrides for everything instead of only for the
+    # appropriate instance/class of the property.
+    if self.class != Puppet::ResourceApi::Property
+      # Define class method insync?(is) if the custom_insync feature flag is set
+      if referrable_type&.type_definition&.feature?('custom_insync')
+        def_custom_insync?
+        if @attribute_name == :rsapi_custom_insync_trigger
+          @change_to_s_value = 'Custom insync logic determined that this resource is out of sync'
+        end
+      # Define class method insync?(is) if the name is :ensure and custom_insync feature flag is not set
+      elsif @attribute_name == :ensure
+        def_ensure_insync?
+      end
+    end
+
     # Pass resource to parent Puppet class.
     super(**resource_hash)
   end
@@ -69,8 +86,51 @@ class Puppet::ResourceApi::Property < Puppet::Property
   # method overloaded only for the :ensure property, add option to check if the
   # rs_value matches is. Only if the class is child of
   # Puppet::ResourceApi::Property.
-  def def_insync?
+  def def_ensure_insync?
     define_singleton_method(:insync?) { |is| rs_value.to_s == is.to_s }
+  end
+
+  def def_custom_insync?
+    define_singleton_method(:insync?) do |is|
+      provider    = @referrable_type.my_provider
+      context     = @referrable_type.context
+      should_hash = @resource.rsapi_canonicalized_target_state
+      is_hash     = @resource.rsapi_current_state
+      title       = @resource.rsapi_title
+
+      raise(Puppet::DevError, 'No insync? method defined in the provider; an insync? method must be defined if the custom_insync feature is defined for the type') unless provider.respond_to?(:insync?)
+
+      provider_insync_result = provider.insync?(context, title, @attribute_name, is_hash, should_hash)
+
+      case provider_insync_result
+      when nil
+        # If validating ensure and no custom insync was used, check if rs_value matches is.
+        return rs_value.to_s == is.to_s if @attribute_name == :ensure
+        # Otherwise, super and rely on Puppet::Property.insync?
+        super(is)
+      when TrueClass, FalseClass
+        # When returning an actual boolean, use default change messaging
+        return provider_insync_result
+      when String
+        # When returning a string, set the change message to the result and return false
+        @change_to_s_value = provider_insync_result
+        return false
+      else
+        # When returning anything else, warn about a non-idiomatic return and return false
+        context.warning("Custom insync for #{@attribute_name} returned a #{provider_insync_result.class} instead of a String")
+        return false
+      end
+    end
+
+    define_singleton_method(:change_to_s) do |current_value, newvalue|
+      # As defined in the custom insync? method, it is sometimes useful to overwrite the default change messaging;
+      # The enables a user to return a more useful change report than a strict "is to should" report.
+      # If @change_to_s_value is not set, Puppet writes a generic change notification, like:
+      #   Notice: /Stage[main]/Main/<type_name>[<name_hash>]/<property name>: <property name> changed <is value> to <should value>
+      # If #change_to_s_value is *nil* Puppet writes a weird empty message like:
+      #   Notice: /Stage[main]/Main/<type_name>[<name_hash>]/<property name>:
+      @change_to_s_value || super(current_value, newvalue)
+    end
   end
 
   # puppet symbolizes some values through puppet/parameter/value.rb
--- a/spec/acceptance/custom_insync_spec.rb
+++ b/spec/acceptance/custom_insync_spec.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'spec_helper'
 require 'tempfile'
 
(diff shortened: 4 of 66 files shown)


lib/puppet/resource_api/property.rb
```diff
@@ -16,9 +16,16 @@ def initialize(type_name, data_type, attribute_name, resource_hash)
...
+      if @attribute_name == :ensure
+        def_insync?
+      elsif @custom_insync
+        def_custom_insync?
+      end
+    end
+
     # Pass resource to parent Puppet class.
-    super(**resource_hash)
+    super(resource_hash)
```
michaeltlombardi, line 28: I think this was an unintentional change I can revert.

michaeltlombardi:
I haven't added the new tests yet but it looks like existing tests are failing on finding the servername fact. 🤔

DavidS (approved):

lib/puppet/resource_api/property.rb
```diff
@@ -69,10 +83,25 @@ def rs_value
...
 
+  def def_custom_insync?
+    define_singleton_method(:insync?) do |is|
+      provider    = @referrable_type.my_provider
+      context     = @referrable_type.context
+      should_hash = @resource.rsapi_canonicalized_target_state
+      is_hash     = @resource.rsapi_current_state
+      title       = @resource.rsapi_title
+
+      provider_insync_result = provider.insync?(context, title, @attribute_name, is_hash, should_hash) if provider.respond_to?(:insync?)
```
DavidS, line 98: If the custom resource has declared 'custom_insync' feature, it MUST provide the `insync?` method, no need to check (although a friendly error message might be useful).

lib/puppet/resource_api/property.rb
```diff
@@ -69,10 +83,25 @@ def rs_value
...
+      return provider_insync_result unless provider_insync_result.nil?
```
DavidS, line 99: I'm sorely tempted to require that this should require only one of `[true, false, nil]` to be returned to avoid accidental leakage of values from ruby's return-less returns.

Either way, the resource api should drop any reference to provider-specified data here and cast explicitly to bool: `return !!provider_insync_result unless ...`

DavidS (changes requested):

DavidS (approved):

spec/fixtures/test_module/lib/puppet/provider/test_custom_insync/test_custom_insync.rb
```diff
@@ -0,0 +1,72 @@
...
+        return true
+      end
+    elsif !should_hash[:version].nil? && should_hash[:version].match?(%r{^\D\s+})
+      context.notice("Checking a custom version string")
+      return Gem::Dependency.new('', should_hash[:version]).match?('', is_hash[:version])
+    end
+  end
+
+  def insync?(context, name, attribute_name, is_hash, should_hash)
+    context.notice("Syncing #{attribute_name}")
```
DavidS, line 54:
```suggestion
    context.notice("Checking whether #{attribute_name} is out of sync")
```

DavidS:
One thing I noticed in trying this out (https://github.com/puppetlabs/opv/commit/a7f60f2af0d55f70d3c231c4010ef0cb48e8f776) is that the type needs to have a property to trigger an insync-check. I expect that to be surprising, confusing, easily missed and dangerous to folks using this feature.

At least this change should have a check that refuses to have a type with custom_insync feature and no property. I would prefer it if this condition is detected that the rsapi injects a hidden property just to force the check being called. This needs a bit more thought and care around how to name that to avoid conflicts and more confusion, but I think it is doable.

michaeltlombardi:
> I would prefer it if this condition is detected that the rsapi injects a hidden property just to force the check being called. This needs a bit more thought and care around how to name that to avoid conflicts and more confusion, but I think it is doable.

@DavidS from a developer writing a resource, is this more or less unexpected than an error message informing the developer that use of custom insync requires an attribute without the namevar, read_only, or parameter behaviors? I'm not sure this is a big deal (since we can just explain it in the docs) but I'm equally unsure if this behavior would be more surprising than raising a `DevError` - though it would give a reusable pattern for resources which do not have a property (also unsure how common that is outside of opv).

From a naming perspective, I think `property_to_trigger_custom_insync` might be both clear and unlikely to collide with anything else. I'm peeking around in the code to see if I can spot a sensible place to add the hidden attribute.

DavidS:
@michaeltlombardi on the risk of bikeshedding, `rsapi_custom_insync_trigger` might be crisper. Either way you call it, I'm not gonna keep you from doing more work for better outcomes :-D

alexjfisher:
I've given this a go with a provider I wrote last month.  Works great! (Well, it would if the rundeck API wasn't so hopeless!)

michaeltlombardi:
@alexjfisher, @DavidS I added a new handler for cases where you want to use custom insync on a resource that doesn't have any insyncable properties; it creates a new hidden attribute, `rsapi_custom_insync_trigger`, which will ensure that `insync?` gets called during a Puppet apply run; you can then insert whatever logic you want into the `insync?` method to validate state.

There's still some work and noodling left to do with regard to reporting and visibility, I'm not sure _how_ hidden we can make the property - as far as I can tell, if `insync?` returns `false`, you're going to get the Puppet reporting around `changed rsapi_custom_insync_trigger...` but there's probably some doable work to pull it from as many places as possible.

michaeltlombardi:
@joshcooper I _think_ this is in a state for code review at this time.

DavidS (approved):

lib/puppet/resource_api/parameter.rb
```diff
@@ -13,7 +13,7 @@ class Puppet::ResourceApi::Parameter < Puppet::Parameter
   # @param attribute_name the name of attribue of the parameter
   # @param resource_hash the resource hash instance which is passed to the
   # parent class.
-  def initialize(type_name, data_type, attribute_name, resource_hash)
+  def initialize(type_name, data_type, attribute_name, resource_hash, _referrable_type)
```
joshcooper, line 16: Is `Puppet::ResourceApi::Parameter.new` public? If so, then this is a breaking change and I think you'll want to provide a default value for `_referrable_type`.

lib/puppet/resource_api/property.rb
```diff
@@ -11,12 +11,29 @@ class Puppet::ResourceApi::Property < Puppet::Property
   # @param attribute_name the name of attribue of the property
   # @param resource_hash the resource hash instance which is passed to the
   # parent class.
-  def initialize(type_name, data_type, attribute_name, resource_hash)
+  def initialize(type_name, data_type, attribute_name, resource_hash, referrable_type)
```
joshcooper, line 14: Same question as above, but for `Puppet::ResourceApi::Property.new`?

lib/puppet/resource_api/property.rb
```diff
@@ -69,10 +86,40 @@ def rs_value
...
+      context     = @referrable_type.context
+      should_hash = @resource.rsapi_canonicalized_target_state
+      is_hash     = @resource.rsapi_current_state
+      title       = @resource.rsapi_title
+
+      raise(Puppet::DevError, 'No insync? method defined in the provider; an insync? method must be defined if the custom_insync feature is defined for the type') unless provider.respond_to?(:insync?)
+      provider_insync_result = provider.insync?(context, title, @attribute_name, is_hash, should_hash)
+      # if nil, super and rely on Puppet::Property.insync?
+      # if true/false, return that and use the default change messaging
+      # if string, set value of custom change reporting and treat as false
```
joshcooper, line 105: It's surprising to me that the `insync?` predicate method can return so many different values, and due to the lack of types in ruby, a provider could accidentally return string `"true"` instead of boolean `true`, and be treated as false. Similar confusion can result if the provider returns symbol `:true`:

```
irb(main):006:0> :true == true
=> false
```


It might be better to return two values instead:

```ruby
result, message = provider.insync(...)
```

where `result` is always `true/false` and `message` is always a non-nil string when `result` is false.

```ruby
irb(main):001:1* def insync?
irb(main):002:1*   true
irb(main):003:0> end
=> :insync?
irb(main):004:0> result, message = insync?
=> true
irb(main):005:0> message
=> nil
```

Ideally I think it would be better to return an `InsyncResult` object, but that might be a breaking API change for existing RSAPI providers?

lib/puppet/resource_api/property.rb
```diff
@@ -69,10 +86,40 @@ def rs_value
...
+      @change_to_s_value = provider_insync_result if provider_insync_result.is_a?(String)
+      return provider_insync_result == true unless provider_insync_result.nil?
+
+      super(is)
```
joshcooper, line 109: It seems odd to me that the provider would define `insync?` but return `nil` to get the "default behavior", which they would have gotten if they didn't define `insync?` in the first place. Can we just say if you define `insync?`, then the provider has to handle it (which might include calling `super(is)` on the underlying `Puppet::Property`)?

spec/acceptance/custom_insync_spec.rb
```diff
@@ -0,0 +1,249 @@
...
+        expect(puppet_apply_stdout).to match %r{Error:}
+        expect(puppet_apply_stdout).to match %r{change from 'FooBar' to 'RaiseError' failed}
+      end
+    end
+
+    context 'when handling subset array comparisons' do
+      let(:manifest) { "test_custom_insync { example: ensure => 'present', some_array => #{test_value} }" }
+
+      context 'when the should array is an exact match for the actual value' do
+        let(:test_value) { %w[a b] }
```
joshcooper, line 24: Does RSAPI behave the same as "raw" puppet where order is significant (ie it's an array and not a set), if so maybe add a test? Puppet can support set-behavior, but it's up to the type/provider to ignore order, for example the `members` of a Windows `group` are not order dependent: https://github.com/puppetlabs/puppet/blob/3aecb85fd55235d0d5f98170d3268a756f000a17/lib/puppet/provider/group/windows_adsi.rb#L33-L37

lib/puppet/resource_api/property.rb
```diff
@@ -69,10 +86,40 @@ def rs_value
...
+  def def_custom_insync?
+    define_singleton_method(:insync?) do |is|
+      provider    = @referrable_type.my_provider
+      context     = @referrable_type.context
+      should_hash = @resource.rsapi_canonicalized_target_state
+      is_hash     = @resource.rsapi_current_state
+      title       = @resource.rsapi_title
+
+      raise(Puppet::DevError, 'No insync? method defined in the provider; an insync? method must be defined if the custom_insync feature is defined for the type') unless provider.respond_to?(:insync?)
+      provider_insync_result = provider.insync?(context, title, @attribute_name, is_hash, should_hash)
```
joshcooper, line 102: Since `insync?` is an extension point, it seems likely that it could raise. Is any additional exception handling needed here to ensure the resource is marked as failed, a `Puppet::Resource::Status` created for the resource in the report, the error is captured in the log, including backtrace info (when running with `--trace`)?

lib/puppet/resource_api.rb
```diff
@@ -166,6 +182,21 @@ def to_resource_shim(resource)
...
+      # insync check; there's no point in specifying it in a manifest as it can only have one
+      # value; it cannot be specified in a type definition as it should only exist in this case.
+      if type_definition.feature?('custom_insync') && type_definition.insyncable_attributes.empty?
+        custom_insync_trigger_options = {
+          type: 'Enum[do_not_specify_in_manifest]',
+          desc: 'A hidden property which enables a type with custom insync to perform an insync check without specifying any insyncable properties',
+          default: 'do_not_specify_in_manifest',
+        }
+
+        type_definition.create_attribute_in(self, :rsapi_custom_insync_trigger, :newproperty, Puppet::ResourceApi::Property, custom_insync_trigger_options)
```
joshcooper, line 197: Creating a hidden property seems a bit of a hack. Could RSAPI instead define a [top-level `validate` for the type](https://github.com/puppetlabs/puppet-specifications/blob/master/language/resource_types.md#validate), and check for missing insync properties there? For example `self[<name>]` will be `nil` if a desired value wasn't specified in the manifest.

michaeltlombardi: The hidden property is meant to address types which intentionally do _not_ have any insyncable properties - this requirement was originally discovered during exploration for the Puppet Native Testing resources, which don't _ensure_ any state, just validate it; rather than have authors of test resources (or any resource which only has parameters, read only properties, and/or init properties) specify a property which has a default and is never specified in a manifest, rely on this known and documented behavior for traditionally un-insyncable resources.

lib/puppet/resource_api/property.rb
michaeltlombardi: I can add some testing to validate behavior for when `insync?` raises and if it doesn't work as expected, we can refactor to be sure that it does.

spec/acceptance/custom_insync_spec.rb
michaeltlombardi: It does, the missing test is an oversight, I'll get that added - the `test_custom_insync` resource in the fixtures module implements this but tests don't actually validate it.

lib/puppet/resource_api/property.rb
michaeltlombardi: The idea for the DevX here was to require as little additional handling in the provider; _only_ add handling for things which need custom insync and otherwise `super` back to the `Puppet::Property.insync?` method. The proposed UX here would require users to _always_ specify the `else` block _unless_ all properties of the resource require custom insync logic, which I suspect would be very rare.

joshcooper: Sure that makes sense. So returning true/false means insync handled it, otherwise nil means do the default.

lib/puppet/resource_api/property.rb
joshcooper: For this issue, I'm fine if `insync?` returns tri-state true/false/nil. But it seems strange to force the implementation to return changed message when the resource is insync. See my comment about `change_to_s` below.

lib/puppet/resource_api/property.rb
```diff
@@ -69,10 +86,40 @@ def rs_value
...
+    end
+
+    define_singleton_method(:change_to_s) do |current_value, newvalue|
+      # As defined in the custom insync? method, it is sometimes useful to overwrite the default change messaging;
+      # The enables a user to return a more useful change report than a strict "is to should" report.
+      # If @change_to_s_value is not set, Puppet writes a generic change notification, like:
+      #   Notice: /Stage[main]/Main/<type_name>[<name_hash>]/<property name>: <property name> changed <is value> to <should value>
+      # If #change_to_s_value is *nil* Puppet writes a weird empty message like:
+      #   Notice: /Stage[main]/Main/<type_name>[<name_hash>]/<property name>:
+      @change_to_s_value || super(current_value, newvalue)
```
joshcooper, line 119: It seems klunky for the `insync?` method to pass back a "changed" message when [elsewhere in the resource-api docs we say](https://puppet.com/docs/puppet/7/about_the_resource_api.html#resource_api_runtime_environment-logging-reporting-utilities)

> The provider needs to signal changes, successes, and failures to the runtime environment. The context is the primary way to do this. It provides a structured logging interface for all provider actions. 

Couldn't the provider just override the change_to_s behavior, for properties that need it, such as a password:

```ruby
def set(context, changes)
  changes.each do |change|
    if change[:name] == 'password'
      context.attribute_changed(change[:title], change[:name], old, new, "Changed [redacted] to [redacted]")
    else
      context.attribute_changed(change[:title], change[:name], old, new) # use default message
     end
  end
  do_something unless noop?
end
```

lib/puppet/resource_api/parameter.rb
DavidS:
```suggestion
  def initialize(type_name, data_type, attribute_name, resource_hash, _referrable_type = nil)
```

I didn't consider it to be public, but you're right, it is good practice to treat it like that.

lib/puppet/resource_api/property.rb
DavidS:
```suggestion
  def initialize(type_name, data_type, attribute_name, resource_hash, referrable_type = nil)
```
also requires changes below to assume that the type is not `custom_insync`-able if no referrable_type is passed.

lib/puppet/resource_api/property.rb
```diff
@@ -11,12 +11,29 @@ class Puppet::ResourceApi::Property < Puppet::Property
...
+
+    # Do not want to define insync on the base class because
+    # this overrides for everything instead of only for the
+    # appropriate instance/class of the property.
+    if self.class != Puppet::ResourceApi::Property
+      # Define class method insync?(is) if the name is :ensure
+      if @attribute_name == :ensure
+        def_ensure_insync?
+      # Define class method insync?(is) if the custom_insync feature flag is set
+      elsif referrable_type.type_definition.feature?('custom_insync')
```
DavidS, line 29:
```suggestion
      elsif referrable_type&.type_definition&.feature?('custom_insync')
```

This should do it.

DavidS:
While investigating some of Josh's comments, I found a bug:

If a provider is `custom_insync` and only has a `ensure` property, `type_definition.insyncable_attributes` at https://github.com/puppetlabs/puppet-resource_api/pull/285/files#diff-0ec370497b0c3e355ba944465e2b20d377290f6b2c8d0a59175b0bcb1b3abde7R190 is not `empty?`, but because of https://github.com/puppetlabs/puppet-resource_api/pull/285/files#diff-ea48f8c743251fe388902613b27bc5c99a13fad74632d5eabf47e0ea4fcae52cR26 , and calling `def_ensure_insync?`, the provider's `insync?` method is not called.

I think that resource api should call `insync?` also for `ensure` and fall back to the custom check in `def_ensure_insync?` on `nil`.

lib/puppet/resource_api/property.rb
DavidS: Your assessment is spot on. This is not great. Sadly, the type/provider is not being called in a way that this would actually be implementable.

Puppet first sets all property values that need it, based on `insync?` logic, including calling into `change_to_s` and emits log messages accordingly. Only later `set` (via `flush`) is called.

I mean, I could imagine decoupling the two implementations completely and calling `set` the first time `insync?` is called, store all change notifications and other logs and dolly them out as puppet catches up with reality, but that sounds like a recipe for even more magic and disaster (where log-messages, actions and failures are entirely disconnected from each other).

lib/puppet/resource_api/property.rb
DavidS: I agree that the current return value handling should be more strict. As it stands values that are not `true` and not `is_a?(String)` could slip through.

I'm undecided whether anything other than literal `true` should cause a change notification exposing the return value or whether return values not in the set of `true`, `false`, `nil` and `is_a?(String)` should cause an error like https://github.com/puppetlabs/puppet-resource_api/blob/dab8aa965899af8b30095445dab73f850779f8d2/lib/puppet/resource_api.rb#L354-L361. I'm leaning more towards the exposing approach as it is easy to catch in visual inspection of puppet output (what people commonly do to test their code), and we set `strict` to the wrong level in the default templates for unit testing (https://github.com/puppetlabs/pdk-templates/blob/377df33b66ce9db9bdd55a1b6b7d3bf5ec887b39/config_defaults.yml#L634)

Independent of that, I also like the idea of the two-value return approach, specifically because it allows the additional possibility of returning `[nil, 'change message']` for values that use the default comparison, but require a custom change message. Even in this case, a more strict approach for checking the return types should be used.

PS: Specifically to the examples of `:true` and `"true"`: I'd dearly love to enable `Lint/BooleanSymbol` rubocop, which traditionally has been turned off because of the usage of `:true` in low-level APIs. In Resource API providers, folks never should need to use `:true` or `:false`.

lib/puppet/resource_api/property.rb
joshcooper: > Puppet first sets all property values that need it, based on insync? logic, including calling into change_to_s and emits log messages accordingly

I don't think that's true. The puppet type/provider API calls `change_to_s` during the `sync` phase not `insync?` This can be seen in the backtrace:

```
--> #0  Puppet::Property.change_to_s(current_value#Symbol, newvalue#String) at /home/josh/work/puppet/lib/puppet/property.rb:204
    #1  Puppet::Transaction::ResourceHarness.sync(event#Puppet::Transaction::Event, param#Puppet::Type::Notify::Message, current_value#Symbol, audit_message#String) at /home/josh/work/puppet/lib/puppet/transaction/resource_harness.rb:247
```

spec/acceptance/custom_insync_spec.rb
joshcooper: cool thanks!

lib/puppet/resource_api/property.rb
DavidS: I believe you mean https://github.com/puppetlabs/puppet/blob/3aecb85fd55235d0d5f98170d3268a756f000a17/lib/puppet/transaction/resource_harness.rb#L247 , right? 
This is exactly when `param.sync` (L241) is called and is only called itself if `param.safe_insync?` (L125) returns false.

All of rsapi's action happens in `flush` (https://github.com/puppetlabs/puppet/blob/3aecb85fd55235d0d5f98170d3268a756f000a17/lib/puppet/transaction/resource_harness.rb#L25) after all params have been synced and change log messages already emitted (L21, L89) - and with no way to suppress the change message.

Also note that `flush` only is called if there is at least one change on the resource (L23) (and there is no `noop` in play).

Altogether I don't see a way to use the context as a conduit for this messaging without a significantly larger change to how puppet applies transactions.

lib/puppet/resource_api/property.rb
```diff
@@ -99,14 +99,28 @@ def def_custom_insync?
...
+        return provider_insync_result
+      when String
+        # When returning a string, set the change message to the result and return false
+        @change_to_s_value = provider_insync_result
+        return false
+      else
+        # When returning anything else, warn about a non-idiomatic return, set the change message, and return false
+        context.warning("Custom insync for #{@attribute_name} returned a #{provider_insync_result.class} instead of a String")
+        @change_to_s_value = "#{@attribute_name} changed #{is_to_s(is)} to #{provider_insync_result.inspect}"
+        return false
```
DavidS, line 122: if we're already providing a `warning` here, we might want to fall back to the default message instead? Given that we consider this an invalid use of the API, it's gonna be awkward either way.

michaeltlombardi: I'm fine with that, I think it works out to be functionally the same anyway.

lib/puppet/resource_api/property.rb
joshcooper: Ah I was confused when you said "based on insync? logic", as I took that literally to mean `provider.insync?` calls into `change_to_s`. But I see you're saying in general the per-property insync/sync logic calls into `change_to_s`, and that happens before flush.

lib/puppet/resource_api/property.rb
joshcooper: Just to reiterate, I'd really like to avoid the magic that "boolean true means success, but string means failure and is the message". I think it's better to be explicit about what is control (true/false/nil) vs data (message).

binford2k:
What is this PR blocked on? I'd like to get it merged real soon (today or early next week) so we can get the DSC work unblocked.

Is the API question something that we can deploy temporarily as an undocumented private API and iterate on with the DSC modules?

michaeltlombardi:
@binford2k I have just cleaned up the commits in the PR in case we can/want to merge today and otherwise to limit the noise if further refactoring is required. Per prior conversations with you, I'm fine with merging as is and "soft launching" with any future changes in implementation to be made and merged after the fact.

michaeltlombardi:
Also worth noting: this is a blocker for Puppet Native Testing as well.

joshcooper:
I'm not sure what "soft launch" means?

The one outstanding issue from my perspective is the the API for the `insync?` method. I don't think we should mix the result (true/false/nil) with the message (String). Since it's a public API it's important to get that right before people start writing extensions that use it.

binford2k:
> Since it's a public API it's important to get that right before people start writing extensions that use it.

This is the part we were discussing. If I understand right, the API change won't require changes in existing code. We were proposing treating the change as an undocumented private API consumed only by the DSC modules. As we iterate on it, the pwshlib gem is the only consumer and @michaeltlombardi has committed to doing the work to account for API changes.

I hate doing something so hacky, but we have a publicly messaged timeline and need to get this shipped ASAP

lib/puppet/resource_api/property.rb
michaeltlombardi: Testing added and behavior validated.

DavidS: tests are in https://github.com/puppetlabs/puppet-resource_api/pull/285/files#diff-08381f7c9c76cb719d8c29f2553a09ea03a0505c24fac62b51d03b6af50c6b5fR13

DavidS: :rofl: 5 seconds apart :+1:

DavidS (approved):

michaeltlombardi:
@DavidS, @joshcooper - latest commit refactors the `insync?` implementation to expect both a result and message and adds some safety handling. Tests are updated to validate this behavior.

I _think_ this addresses the last concerns?

joshcooper:
Wanted to add some context that Puppet < 3 had nearly this same problem, the `Puppet::Util::Execution.execute` method only returned a string. So in order get the child exit status, you had to rely on global state `$CHILD_STATUS`. We implemented a terrible hack -- have providers return a `ProcessOutput` that extended `String`: https://github.com/puppetlabs/puppet/commit/7b2cc74b73. All of that could have been avoided if the `execute` method returned an `ExecutionResult` object instead of a `String`. I didn't implement the original `execute`, but wanted to use that as a real world example of what not to do.

About https://github.com/puppetlabs/puppet-resource_api/pull/285/commits/2eeff5767c3d90edbb0e29e1b1bf76f32871982a, since the API is brand new, why not make symbol `:true|:false` and string `"true|false"` be errors? That way there's no confusion about what custom providers are supposed to return.

michaeltlombardi:
@joshcooper that makes sense, I can fix that commit either to warn (keeping it in line with the `else` clause) or error (and convert the else clause to do the same), raising a DevError as the issue is in the provider design, not the runtime.

michaeltlombardi:
@joshcooper, @DavidS, updated the implementation to raise a `DevError` on invalid result return instead of warn and treat as false.

joshcooper (approved):

PR #292, opened 2021-12-13
Support ensure parameter with Optional data type

seanmil:
Add support for specifying the ensure parameter as an
Optional Enum data type.

This permits resource properties of types specified this way
to be managed but only if the resource is already present and
cleanly skipped otherwise.

This is a behavior already present in many traditional Puppet
resource types and this extends the option to the Resource API,
easing transition of existing resource types to the Resource API.


michaeltlombardi (approved):

michaeltlombardi:
LGTM! Merging despite the codecov misbehaving.

Issue #231, opened 2019-09-30
Add `default:` for `connection_info` attributes on transports


DavidS:
## Use Case
Optional attributes for transport `connection_info` often have default values. Currently this requires calling out the default value in the desc string and defining a fallback calculation in code, leading to duplication at best and inconsistency at worst.

## Describe the Solution You Would Like
Define a `default:` key equivalent to the `attributes` in types.

PR #293, opened 2021-12-20
(GH-231) Add default to transport attributes

seanmil:
Add the ability to specify a default to transport connection_info
attributes which will be used in the event no value for the given
connection_info attribute is provided.


michaeltlombardi (approved):
LGTM, thanks for the contribution!

PR #294, opened 2022-01-11
Add array support to autorequire variable expansion

seanmil:
Add support for autorequire-family variable expansion referencing
parameters with array values. Also, remove any nil values from
results.


michaeltlombardi (approved):

michaeltlombardi:
Merging as the improvement and tests seem sufficient even though codecov is misbehaving.

PR #302, opened 2022-10-25
SimpleProvider name fixes

seanmil:
The SimpleProvider has some support for composite namevars, but it is incomplete and only works in scenarios where the :is value is supplied. This PR adds more comprehensive unit tests which cover scenarios where :is is not passed and fixes these issues. As a bonus, the SimpleProvider now works for types with any namevar (it no longer is required to be `name` for the single-valued case).  Some additional code cleanups are included.


LukasAud:
@seanmil Apologies for the rebase, I need to make sure this worked on latest.

LukasAud (approved):
LGTM

Issue #303, opened 2022-10-27
Resource purging does not work with simple_get_filter


seanmil:
## Describe the Bug

For types with the `simple_get_filter` feature it is seemingly not possible to perform resource purging with either the `resources` type or the `purge` type from crayfishx/purge.

When attempting to purge types with `simple_get_filter` it seems that the existing-but-unmanaged resources are never found via the type instances() method, therefore preventing them from being purged.

Removing the `simple_get_filter` flag is sufficient to restore resource purging to a working state.

## Expected Behavior

Types using `simple_get_filter` should support instantiation, even of unknown resources, via instances() OR a warning/error should be thrown.  The user of a given Resource API-based resource should not have to be aware of the implementation details to figure out if resource purging will work.

## Environment
 - Version Puppet 7.18.0

seanmil:
So, doing a bit more digging, per the [Puppet Specifications on the Resource API](https://github.com/puppetlabs/puppet-specifications/blob/master/language/resource-api/README.md#provider-feature-simple_get_filter):

> Some resources are very expensive to enumerate. The provider can implement simple_get_filter to signal extended capabilities of the get method to address this. The provider's get method will be called with an array of resource names, or nil. The get method must at least return the resources mentioned in the names array, but may return more than those. If the names parameter is nil, all existing resources should be returned. The names parameter defaults to nil to allow simple runtimes to ignore this feature.

However, the resource API doesn't pass `nil`, it passes `[]` https://github.com/puppetlabs/puppet-resource_api/blob/main/lib/puppet/resource_api.rb#L234

I think switching `[]` to `nil` here should be sufficient.

LukasAud:
Issue potentially addressed by #304

LukasAud:
Hi @seanmil, can you confirm whether this issue was addressed by the aforementioned Pull Request? If so, we would like to close this thread.

jordanbreen28:
inline with above comments. please re-open if necessary

PR #304, opened 2022-10-28
Fix instances method with simple_get_filter

seanmil:
Per issue #303, the Puppet Specification for Resource API indicates that a provider get() method when simple_get_filter is enabled should return all instances when the second/names parameter is set to 'nil'.

Currently, if a Resource API provider is implemented per specification and uses the simple_get_filter feature it will never return any resources for the `instances()` method, which prevents things such a resource purging from working.

### DIFF
--- a/lib/puppet/resource_api.rb
+++ b/lib/puppet/resource_api.rb
@@ -248,7 +248,7 @@ module Puppet::ResourceApi
         provider(type_definition.name)
 
         initial_fetch = if type_definition.feature?('simple_get_filter')
-                          my_provider.get(context, [])
+                          my_provider.get(context, nil)
                         else
                           my_provider.get(context)
                         end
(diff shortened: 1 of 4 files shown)


seanmil:
So, after sleeping on this, I'm not sure that changing the behavior in the code to match the documentation is the right change. It could be a functionally breaking change for someone who wrote code based on the actual implementation and not the documented behavior.

Instead maybe it makes sense to update the documentation. I put up a PR for the Puppet Specifications at https://github.com/puppetlabs/puppet-specifications/pull/155 but the Puppet docs on the docs website would also need to be fixed.

LukasAud:
Closing and re-opening to re-kick testing.

LukasAud (approved):
LGTM

LukasAud:
@seanmil Apologies for the rebase, I need to make sure this worked on latest.

Issue #305, opened 2022-10-31
Provider get() is called too many times


seanmil:
## Use Case

The get() method on a provider is currently called too many times. The operations performed by the get() method can often be somewhat time consuming (e.g. slow command, slow API). The Resource API specification at https://github.com/puppetlabs/puppet-specifications/blob/master/language/resource-api/README.md acknowledges that when it states that calls to get() should be minimal:

* "The runtime environment calls get with a minimal set of names, and keeps track of additional instances returned to avoid double querying. To gain the most benefits from batching implementations, the runtime minimizes the number of calls into get." from https://github.com/puppetlabs/puppet-specifications/blob/master/language/resource-api/README.md#provider-feature-simple_get_filter
* "A transaction will usually call get once" from: https://github.com/puppetlabs/puppet-specifications/blob/master/language/resource-api/README.md#runtime-environment

The problem however is that, in practice, the calls to get are not always minimal.

Here are the problems with the current implementation:

* The instances() method is called in primarily (only?) two scenarios: 1) "puppet resource" is used. and 2) By resources which use "resource generation" (probably to support purging: see "resources" and "crayfishx/purge".  instances() attempts to cache results with the "cache_current_state" method, but that caches state into each resource instance and those are not preserved between calls to instances() - so they are not available to other instances() calls nor to Puppet for the normal apply transaction.
* When the provider does not use "simple_get_filter" the state is retrieved in refresh_current_state per resource instance, however, only the results for the sought resource are saved, discarding everything else returned by get. get() is called this way once per resource: https://github.com/puppetlabs/puppet-resource_api/blob/main/lib/puppet/resource_api.rb#L256

The case where get() is called (in a type which does not support "simple_get_filter") and returns all resources it is particularly bad because that is called to enumerate all the resources perhaps hundreds (or thousands) of times, for a use-heavy resource, with all the results but one discarded.

## Describe the Solution You Would Like

Per the Resource API specification, the Resource API should actually attempt to minimize the calls to get(). Once a resource is retrieved, it should not be retrieved a second time.

The Resource API should establish and manage a cache which persists for the duration of a "puppet resource", "puppet apply" or single "puppet agent" transaction which contains any values fetched from the provider get().

## Describe Alternatives You've Considered

[...]

PR #306, opened 2022-10-31
Provider get() call optimization

seanmil:
This PR adds a caching laying which caches the results returned by the provider's get() method to try to actually minimize the calls to get(), per the Puppet Resource API specifications.


seanmil:
This is one possible way to address #305, but after a lot of consideration, I feel it is the best approach. The complexity should be kept away from the implementer and given Puppet's current design, I believe this approach will do the trick.

To anticipate the one big question I expect everyone will have: When Puppet is running as a daemonized agent this still receives a new cache per run. It seems that the type and provider are instantiated as part of the transaction during the Puppet run, following the fork(). When Puppet is running as "puppet resource" or "puppet apply" it is a one-shot mode and so it will be a one-transaction cache regardless. I confirmed this in my testing, but I am unsure as to how to write a test for it. Nevertheless, in all of my testing this is a massive improvement for environments where get() is a resource/time heavy operation.

fraenki:
This looks super interesting. @joshcooper Could someone please review this PR? Thanks :)

fraenki:
Hey @seanmil, maybe you could create an issue in the [Puppet issue tracker](https://tickets.puppetlabs.com/)? From my experience this may help to get some resource assigned to a PR. :)

LukasAud:
Closing and re-opening PR to re-kick testing run. Interested to see if this is green.

LukasAud (approved):
LGTM. Tested branch against Firewall module to ensure nothing breaks outside our CI.

PR #311, opened 2023-01-25
(maint) Update to Mend

cthorn42:
Also patch Ruby 2.5 to use a known good Bundler.


joshcooper (approved):
Looks like branch protection rules need updating, otherwise looks good

PR #316, opened 2023-06-23
(CAT-761) Add custom_generate as a feature

david22swan:
Allows the generate function to be utilized in order to manually create an array of types to be enforced.
The `generate` method is called prior to the resource being managed.
There is a second `eval_generate` method that is called after it has been, which may be good to add at a later date.

### DIFF
--- a/lib/puppet/resource_api.rb
+++ b/lib/puppet/resource_api.rb
@@ -136,6 +136,23 @@ module Puppet::ResourceApi
         @rsapi_canonicalized_target_state
       end
 
+      # Method is used to custom generate resources which are then applied by Puppet
+      def generate
+        # If feature `custom_generate` has been set then call the generate function within the provider and return the given results
+        return unless type_definition&.feature?('custom_generate')
+        should_hash = rsapi_canonicalized_target_state
+        is_hash     = rsapi_current_state
+        title       = rsapi_title
+
+        # Ensure that a custom `generate` method has been created within the provider
+        raise(Puppet::DevError, 'No generate method found within the types provider') unless my_provider.respond_to?(:generate)
+        # Call the providers custom `generate` method
+        rules_resources = my_provider.generate(context, title, is_hash, should_hash)
+
+        # Return array of resources
+        rules_resources
+      end
+
       def rsapi_current_state
         refresh_current_state unless @rsapi_current_state
         @rsapi_current_state
--- a/lib/puppet/resource_api/type_definition.rb
+++ b/lib/puppet/resource_api/type_definition.rb
@@ -36,7 +36,7 @@ module Puppet::ResourceApi
       Puppet::ResourceApi::DataTypeHandling.validate_ensure(definition)
 
       definition[:features] ||= []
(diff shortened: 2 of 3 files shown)


lib/puppet/resource_api.rb
```diff
@@ -136,6 +136,23 @@ def rsapi_canonicalized_target_state
...
+        # If feature `custom_generate` has been set then call the generate function within the provider and return the given results
+        return unless type_definition&.feature?('custom_generate')
+        should_hash = rsapi_canonicalized_target_state
+        is_hash     = rsapi_current_state
+        title       = rsapi_title
+
+        # Ensure that a custom `generate` method has been created within the provider
+        raise(Puppet::DevError, 'No generate method found within the types provider') unless my_provider.respond_to?(:generate)
+        # Call the providers custom `generate` method
+        rules_resources = my_provider.generate(context, title, is_hash, should_hash)
```
joshcooper, line 150: Confusingly puppet's type and provider system supports two styles of generating resources on the agent: `generate` and `eval_generate`, and I assume this is relying on the former?

If a `Puppet::Type::<name>` class implements the `generate` method, then it will be called early on in the agent's catalog application process, before any resources have been managed. This is often the wrong thing to do, because there's no guarantee any of the other resources in the catalog have reached their desired state before *this* resource's `generate` method is called.

The generally more useful method is `eval_generate`, because it is called while *this* resource is being evaluated. So all of its dependencies have already been managed/enforced. For example, the `Puppet::Type::File` class implements `eval_generate`, which is needed for recursive directory operations. We have to use `eval_generate` since all of the parent directories need to be managed with the correct permissions before we try to recurse: https://github.com/puppetlabs/puppet/blob/fde57136980d63bd2a64532ad74f750cc3b73bc7/lib/puppet/type/file.rb#L531

However, if a type implements prefetching, then you may be forced to use `generate`. For example, see https://tickets.puppetlabs.com/browse/PUP-2718?focusedCommentId=839874&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-839874 and our attempts to switch `sshkeys` from `generate` to `eval_generate` and back to `generate` in PUP-11320.

If the resource API doesn't support prefetching, then I would recommend using `eval_generate` instead. It's usually how people assume `generate` works.

david22swan: In this case I think that `generate` is preferred as we actively want the resources being fetched to be overwritten.
By that I mean we are generating delete commands against all active Firewall rules on the chain and then letting the catalog overwrite any that we actually need.
From the way you described `eval_generate` I am not sure in would work in the same way.

david22swan: @joshcooper ??

david22swan:
@joshcooper Have updated the PR to add a `custom_eval_generate` feature as well.
It's pretty much just a clone of the `custom_generate` method as they both just exist as wrappers to call a method defined within the users local provider.

joshcooper (approved):

PR #325, opened 2023-07-19
(PA-5641) Change github bump action and release

tvpartytonight:
The old github action from anothrNick/github-tag-action only used lightweight tags; this caused compatibility problems with the way puppetlabs-jenkins used annotated tags. In order to keep the tags consistent, this commit uses mathieudutour/github-tag-action which supports annotated tags. Since the previous attempt to bump skipped the pre-bumped version, that old tag is still available and can be used to synchronize the version in code and in tag.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -15,15 +15,11 @@ jobs:
         with:
           fetch-depth: '0'
       - name: Bump version and push tag
-        uses: anothrNick/github-tag-action@1.67.0
-        env:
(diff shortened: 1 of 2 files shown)


tvpartytonight:
There is still an issue with the rubygems issue not having permission; once that hurdle is cleared we should be able to merge this change. 🤞

joshcooper (approved):

tvpartytonight:
Removed the bumpiing entirely from the GH action, no longer needed.

PR #352, opened 2024-06-11
Fix broken dependencies

gavindidrichsen:
## Summary

Recent rubocop updates broke the resource_api's CI:

```ruby
➜  puppet-resource_api git:(main) bundle exec rubocop
Error: Property AutoCorrect of cop FactoryBot/CreateList is supposed to be a boolean and contextual is not.
+ set +x
➜  puppet-resource_api git:(main) 
```

Although the fix for `ruby` environments was to change the Gemfile from this:

```ruby
  gem 'rubocop', '~> 1.48.1', require: false
  gem 'rubocop-rspec', '~> 2.19', require: false
```

to this:

```ruby
gem 'rubocop', '~> 1.64.1', require: false
gem 'rubocop-rspec', '~> 3.0', require: false
```

The same could not be applied to the resource_api's `jruby` environments.  This PR adds a workaround to keep both the `ruby` and `jruby` environments passing in CI.

**NOTE**: I also moved some of the commentary about rubocop from the `Gemfile` and into separate design decision documents.  See below for more information.

## For more information

For more information about why the above changes were made, see 
* [docs/adr/0001-test-the-resource-api-against-jruby-as-well-as-ruby-engines.md](https://github.com/puppetlabs/puppet-resource_api/blob/f7fe2e59fc6a72ab213da281b2d98361d47f5fea/docs/adr/0001-test-the-resource-api-against-jruby-as-well-as-ruby-engines.md) and 
* [docs/adr/0002-split-gemfile-dependencies-by-jruby-and-ruby-engines.md](https://github.com/puppetlabs/puppet-resource_api/blob/f7fe2e59fc6a72ab213da281b2d98361d47f5fea/docs/adr/0002-split-gemfile-dependencies-by-jruby-and-ruby-engines.md)

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [x] Manually verified.


.rubocop_todo.yml
```diff
@@ -1,29 +1,108 @@
...
 
+# Offense count: 1
+# This cop supports safe autocorrection (--autocorrect).
+# Configuration parameters: AllowForAlignment, AllowBeforeTrailingComments, ForceEqualSignAlignment.
+Layout/ExtraSpacing:
+  Exclude:
+    - 'spec/puppet/resource_api/transport_schema_def_spec.rb'
+
+# Offense count: 1
+# This cop supports safe autocorrection (--autocorrect).
```
david22swan, line 17: Might be good to try and do the safe autocorrects

david22swan (approved):
Look's ok but would want another pair of eyes before merge due to the ruby split

.rubocop_todo.yml
gavindidrichsen: Hi @david22swan.  I thought about getting this across the line with a green CI and then doing further autocorrects in another PR.  The autocorrects I included in this PR were necessary to keep CI green for the `jruby` environment tests.

PR #366, opened 2025-02-18
(CAT-2235) Troubleshoot failing nightlies

LukasAud:
Currently, CI and nightlies are failing due to a JRuby installation issue. This commit attempts to revert the provisioning ubuntu to an earlier version to troubleshoot the issue.


joshcooper (approved):
We had to make the same change in puppet https://github.com/puppetlabs/puppet-private/commit/2ac1ec0d1b20945480a6ce2e0e7bf8011160f18b

PR #367, opened 2025-02-21
(CAT-2113) Update spec_helper version

LukasAud:
This commit aims to update the severely outdated spec_helper depencency one version at a time until we reach the latest version.

PR #368, opened 2025-03-10
(CAT-2254) Remove puppet 7 / Ruby 2.7 related code

LukasAud:
Following the EOL of Puppet 7, we are updating the codebase to remove Puppet 7 related code. Changes made include the removal of Ruby 2.7 testing in our Actions matrix and Rubocop bumps to newer versions, alongside a new TargetRuby setting and several autocorrections, and some outdated version conditional removals.

This PR will enable #367 to be merged


.github/workflows/ci.yml
```diff
@@ -13,8 +13,6 @@ jobs:
     strategy:
       matrix:
         cfg:
-          - {puppet_version: '7', ruby: '2.7'}
-          - {puppet_version: '7', ruby: 'jruby-9.3.7.0'}
```
joshcooper, line 17: Two potential issues are:

1. We need to maintain compatibility with older agents that haven't upgraded yet. We always want to support upgrading old agents
1. s puppet types and providers may be loaded in puppetserver and JRuby 9.4 in puppet 8 maintains compatibility with MRI 3.1 language features. So resource-api can't use any ruby language features that were added in MRI 3.2. As long as we continue to test with rubocop nightly then it should be ok to drop 

As long as we run rubocop as part of our CI check that that should be ok to remove puppet7 completely

david22swan (approved):

joshcooper (approved):

PR #376, opened 2025-05-21
(packaging) Bump to version '2.0.1' [no-promote]

AriaXLi:
Since puppet-resource_api v2.0.0 was released out of band from a Puppet Core release, it's version needs to be set to the next release, 2.0.1, in preparation for Puppet Core 8.13.0


joshcooper:
I think maybe we want to keep the reference to v2.0.0, let's talk tomorrow.

PR #378, opened 2025-06-05
(CAT-2327) Temporary revert of cache changes

LukasAud:
Following an escalation on resource_api, other teams are experiencing blockers due to changes made to cache behaviour in between 1.9.0 and 2.0.0. We are reverting changes related to cache management made in between those versions for a quick release.

The core of this change is the reversion of the file resource_api.rb to its 1.9.0 state. The goal is to have a hotfix ready for a release if needed. This PR should be reverted after release and main branch code revisited for conflicts with PE behaviour.


LukasAud:
This is the command used for reverting:
`git checkout 1.9.0 -- lib/puppet/resource_api.rb`

LukasAud:
Outdated. No longer relevant.

joshcooper:
@LukasAud we had to revert puppet-agent packages back to resource-api 1.9 due to a regression in the 2.0 version. https://github.com/puppetlabs/puppet-agent-private/commit/d57e065774d0d16c65e522ff59e41345cf05c4cb

Is there another ticket/PR for fixing the regression so that we can unpin in the future?

PR #381, opened 2025-12-12
(MODULES-11462): Implement finish method for class tag inheritance in Puppet::ResourceApi in 1.9.0 branch

span786:
## Summary
Implement the finish method for class tag inheritance within Puppet::ResourceApi.
This method is being overridden to allow reinjection of default tags (such as the class name) back into the Puppet catalog during compilation.

## Additional Context
- [x] The Puppet::ResourceApi currently does not handle metaparams like tags effectively.
- [x] To handle default tags, we needed a point in the catalog-building process where they could be reintroduced. Overriding the finish method provides that opportunity, enabling us to push the tags back into the catalog at the right stage.

## Related Issues (if any)
Due to this limitation, modules such as `puppetlabs-firewall` (version 8.0.0 and later) fail to recognize default tags like the class name. As a result, running `puppet apply <manifest.pp> --tags <classname>` does nothing because the expected tags are missing.

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [x] Manually verified.

### DIFF
--- a/lib/puppet/resource_api.rb
+++ b/lib/puppet/resource_api.rb
@@ -107,6 +107,27 @@ module Puppet::ResourceApi
         super(attributes)
       end
 
+      # Override finish method to ensure scope tags (like class names) are properly inherited
+      # This is called after the resource is added to the catalog and containment is established
+      def finish
+        super if defined?(super)
+        return unless @catalog
+
+        # Use pathbuilder to tag all containing classes
+        # Pathbuilder returns the containment hierarchy; class names appear as plain strings
+        # while other resources have the format "Type[title]"
+        return unless respond_to?(:pathbuilder)
+
+        pathbuilder.each do |container|
+          next unless container.is_a?(String)
+
+          # Classes don't contain '[' or ']' characters, resources do
+          # Classes: "Test::Modules_11462", "Settings"
+          # Resources: "Stage[main]", "Firewall[001 test rule]"
+          tag(container) unless container.include?('[')
+        end
+      end
+
       def name
         title
       end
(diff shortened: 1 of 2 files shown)


span786:
The `puppet-resource_api` gem version `2.0.0` is currently unstable, so `1.9.0` has been tagged for puppet-agent releases.
This PR targets the `1.9.0` tag to allow us to release version `1.9.1` with the changes introduced here, ensuring no breaking changes are included in the puppet-agent release.

skyamgarp (approved):

lib/puppet/resource_api.rb
```diff
@@ -107,6 +107,27 @@ def initialize(attributes)
...
+        # while other resources have the format "Type[title]"
+        return unless respond_to?(:pathbuilder)
+
+        pathbuilder.each do |container|
+          next unless container.is_a?(String)
+
+          # Classes don't contain '[' or ']' characters, resources do
+          # Classes: "Test::Modules_11462", "Settings"
+          # Resources: "Stage[main]", "Firewall[001 test rule]"
+          tag(container) unless container.include?('[')
```
joshcooper, line 127: Leaving a comment for posterity. We've had issues in the path with metaparams not being applied to `Puppet::Type` instances that are generated on the agent at runtime, such as tag, schedule, noop, etc. We added a method https://github.com/puppetlabs/puppet/blob/e227c27540975c25aa22d533a52424a9d2fc886a/lib/puppet/type.rb#L365 to do that and then called the method in places where resources are generated such as when the `User` generates `Puppet::Type::Ssh_authorized_key` instances 
https://github.com/puppetlabs/puppet/blob/e227c27540975c25aa22d533a52424a9d2fc886a/lib/puppet/type/user.rb#L818

I would be curious to understand whether the existing `copy_metaparams` isn't being called? Or if it is being called, but it doesn't walk the containment hierarchy?

One suggestion is to rename the `1_9_0_base` branch to `1.9.x` so it follows our other repo conventions.

PR #390, opened 2026-06-30
(PA-8346) Harden sensitive parameter preservation against regressions

gavindidrichsen:
## Summary

Follow-up to PR #388 which fixed the cleartext leak of sensitive parameters during `Puppet::Resource` → Hash conversion. This PR hardens that fix in three ways:

- **Replace fragile `attributes[:sensitive_parameters]` injection with `set_sensitive_parameters`** — PR #388 injected `:sensitive_parameters` into the attributes hash before `super`, relying on `hash2resource`'s undocumented internal key-deletion to clean up the extra key. This works today but couples to an implementation detail of Puppet core. This PR uses the public `Puppet::Type#set_sensitive_parameters` API after `super` instead, which is both the intended API and resilient to upstream changes in `hash2resource`.

- **Add a re-wrap guard after `canonicalize`** — Providers that implement `canonicalize` receive unwrapped attribute values. If a provider returns a plain string where a `Sensitive`-wrapped value was expected, the sensitive wrapping is silently lost. This PR adds a post-canonicalize re-wrap pass (identical to the existing pre-canonicalize one) so that the `Sensitive` wrapper survives regardless of provider behavior.

- **Add the missing regression test** — PR #388's tests covered the "no sensitive params" case and the Sensitive-wrapping case, but did not test the primary fix path: a `Puppet::Resource` input carrying a non-empty `sensitive_parameters` list with unwrapped values. This PR adds that test, directly asserting `parameter(:secret).sensitive == true` and that the value is re-wrapped as `Sensitive` — the test that would catch a reintroduction of the original bug.

## Test Plan

- [x] Existing specs continue to pass
- [x] New spec `when loading from a Puppet::Resource with sensitive parameters listed` verifies:
  - `parameter(:secret).sensitive` returns `true`
  - The secret value is re-wrapped as `Puppet::Pops::Types::PSensitiveType::Sensitive`

## Related

- Fixes: [PA-8346](https://tickets.puppetlabs.com/browse/PA-8346)
- Follows: #388

### DIFF
--- a/lib/puppet/resource_api.rb
+++ b/lib/puppet/resource_api.rb
@@ -96,6 +96,11 @@ module Puppet::ResourceApi
         # $stderr.puts "B: #{attributes.inspect}"
         attributes = my_provider.canonicalize(context, [attributes])[0] if type_definition.feature?('canonicalize')
 
+        # Re-wrap any sensitive values that canonicalize may have unwrapped
+        sensitives.each do |name|
+          attributes[name] = Puppet::Pops::Types::PSensitiveType::Sensitive.new(attributes[name]) if attributes.key?(name) && !attributes[name].is_a?(Puppet::Pops::Types::PSensitiveType::Sensitive)
+        end
+
         # the `Puppet::Resource::Ral.find` method, when `instances` does not return a match, uses a Hash with a `:name` key to create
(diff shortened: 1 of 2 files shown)


gavindidrichsen:
Closing this.  Can possibly re-open at a later date.

